PR #16, opened 2024-07-08
Add Architecture Diagrams

bolny:
Adds ARCHITECTURE.md.

Github's built in diagram renderer isn't the best but it should get the point across.


ARCHITECTURE.md
```diff
@@ -0,0 +1,50 @@
+# SITE Registry Rebuild Architecture
```
nikhila-aot, line 1: Current Process

- Site Oracle Forms: for internal users to make CRUD operations on contaminated sites
- ESRA: for external users to search for a site and purchase any reports 

New Process

- SITE Rebuild: caters to both internal and external users to replicate the current process

Data Integration/Transformation

We will be doing a one time load of all the data in the legacy oracle database to the new postgres instance as part of the rebuild. Steps taken to achieve this:

- typeorm-model-generator: npm package used to generate typeorm entities from oracle DDL 
- code first approach: to generate postgres backend from the entities using typeorm
- ora2pg: to load data from the oracle legacy database to the new postgres instance

Please note that there will be NO two way sync between the legacy and new databases.

nikhila-aot (changes requested):

ARCHITECTURE.md
bolny: I have updated the diagrams based on your feedback, along with a few new things:

1. I decided to use mermaid flowchart diagrams because mermaid's C4 diagram support is currently pretty poor.
2. I added component diagrams for the frontend and backend

nikhila-aot (approved):

PR #21, opened 2024-07-11
SRS-372: Suspect Land Uses Tab

anton-bcgov:
[SRS-372: Suspect Land Uses Tab](https://epddigitalservices.atlassian.net/browse/SRS-372)

Screenshot 2024-07-25 at 9 07 27 AM


backend/sites/src/app/services/landHistory/landHistory.service.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export class LandHistoryService {
+    constructor(
+        @InjectRepository(LandHistories)
+        private landHistoryRepository: Repository<LandHistories>,
+    ) { }
+
+    async getLandHistoriesForSite(siteId: string): Promise<LandHistories[]> {
+        try {
+            return this.landHistoryRepository.find({ where: { siteId } });
```
nikhila-aot: assign this to a variable and return. Wil need to be used for logging in future.

backend/sites/src/app/resolvers/landHistory/landHistory.resolver.ts
```diff
@@ -0,0 +1,26 @@
...
+        private readonly genericResponseProvider: GenericResponseProvider<LandHistories[]>,
+    ) { }
+
+    @Roles({ roles: ['site-admin'], mode: RoleMatchingMode.ANY })
+    @Query(() => LandHistoryResponse, { name: 'getLandHistoriesForSite' })
+    @UsePipes(new GenericValidationPipe())
+    async getLandHistoriesForSite(
+        @Args('siteId', { type: () => String }) siteId: string,
+    ) {
+        const result = await this.landHistoryService.getLandHistoriesForSite(siteId);
```
nikhila-aot: handle the case when result does not contain data

backend/sites/src/app/entities/landUseCd.entity.ts
```diff
@@ -16,7 +16,7 @@ export class LandUseCd {
   @Column("character varying", { name: "description", length: 60 })
   description: string;
 
-  @OneToMany(() => LandHistories, (landHistories) => landHistories.lutCode2)
+  @OneToMany(() => LandHistories, (landHistories) => landHistories.landUse)
```
nikhila-aot, line 18: isn't this lutCode2?

backend/sites/src/app/entities/landHistories.entity.ts
```diff
@@ -69,6 +69,11 @@ export class LandHistories {
   @JoinColumn([{ name: "lut_code", referencedColumnName: "code" }])
   lutCode2: LandUseCd;
 
+  @Field(() => LandUseCd)
```
nikhila-aot, line 68: there is already a relationship existing in line 68-70. Can we use the same?

backend/sites/src/app/services/landHistory/landHistory.service.ts
```diff
@@ -0,0 +1,18 @@
...
+        } catch (error) {
+            throw error;
+        }
+    }
+}
```
nikhila-aot, line 48: pls include unit tests for resolver and services

nikhila-aot (changes requested):

backend/sites/src/app/resolvers/landHistory/landHistory.resolver.ts
anton-bcgov: Could you elaborate? We return an empty array when there's no data found.

backend/sites/src/app/entities/landHistories.entity.ts
anton-bcgov: yeah, I was going to ask you about this. I think `lutCode2` is not a very descriptive name, but I wasn't 100% confident in removing that relationship altogether, so I created a new one. What do you think? Should we leave `lutCode2` as is, remove it, or straight up use it?

backend/sites/src/app/entities/landUseCd.entity.ts
anton-bcgov: discussing this in another thread

backend/sites/src/app/services/landHistory/landHistory.service.ts
anton-bcgov: That's the plan! I didn't expect you to review a Draft PR 😄

midhun-aot (commented):
Frontend changes looks good.

backend/sites/src/app/entities/landHistories.entity.ts
nikhila-aot: we have to leave it intact, we cannot change any existing column names as they are replicated from the oracle db.

backend/sites/src/app/services/landHistory/landHistory.service.ts
nikhila-aot: the team raises the PR when the task is done, not before that 😄

backend/sites/src/app/resolvers/landHistory/landHistory.resolver.ts
nikhila-aot: example:

if (result.length > 0) {
          return this.genericResponseProvider.createResponse('Participants fetched successfully', 200, true, result);
        }
        else
        {
          return this.genericResponseProvider.createResponse(`Participants data not found for site id: ${siteId}`, 404, false);
        }

backend/sites/src/app/entities/landHistories.entity.ts
anton-bcgov: It's not changing the database columns though, right? This only defines a relationship in the application code, the underlying data still comes from the `lut_code` and `code` DB columns,

nikhila-aot: its redundant code, for the same purpose. Just to have a better column name

anton-bcgov: just so we're on the same page, do you agree that we can remove `lutCode2` relationship if it's safe to do so and prefer a clearer variable name?

anton-bcgov:
Sorry for the unrelated formatting changes that are included in this PR. For some reason, formatter didn't run on #23, and ran instead when I did a merge commit from `dev` branch :shrug:

frontend/site-registry/src/app/components/table/body/TableBody.tsx
```diff
@@ -118,184 +119,191 @@ const TableBody: FC<TableBodyProps> = ({
...
+          tableMode={field.tableMode ?? false}
+          stickyCol={field.stickyCol}
+        />
+      );
+    }
   };
 
   const getValue = (rowIndex: number, propertyName: string) => {
-    return data[rowIndex][propertyName];
+    return get(data[rowIndex], propertyName);
```
midhun-aot, line 392: Please remove the import from loadash as we will introducing an external dependency. May be to check for undefined or null we can use (data[rowIndex][propertyName]) ?? ''

backend/sites/src/app/resolvers/landHistory/landHistory.resolver.ts
```diff
@@ -0,0 +1,55 @@
...
+  constructor(
+    private readonly landHistoryService: LandHistoryService,
+    private readonly genericResponseProvider: GenericResponseProvider<
+      LandHistories[]
+    >,
+  ) {}
+
+  @Roles({ roles: ['site-admin'], mode: RoleMatchingMode.ANY })
+  @Query(() => LandHistoryResponse, { name: 'getLandHistoriesForSite' })
+  // TODO: figure out how this works and if we need GenericValidationPipe here
```
nikhila-aot: is there more logic to be added here ?

backend/sites/src/app/resolvers/landHistory/landHistory.resolver.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    @Args('sortDirection', { nullable: true })
+    sortDirection: SortDirection,
+  ) {
+    const result = await this.landHistoryService.getLandHistoriesForSite(
+      siteId,
+      searchTerm,
+      sortDirection,
+    );
+    if (result.length > 0) {
```
nikhila-aot, line 36: should you check if result is empty before performing result.length ?

backend/sites/src/app/entities/landHistories.entity.ts
nikhila-aot: yes, keep the one with the new name and remove the old one, we do not have to retain redundant relationships.

backend/sites/src/app/resolvers/landHistory/landHistory.resolver.ts
anton-bcgov: `getLandHistoriesForSite` calls typeorm's `getMany`, which should always return an array, so I think we're safe here.

backend/sites/src/app/resolvers/landHistory/landHistory.resolver.ts
anton-bcgov: nope, it's an old comment, I deleted it

backend/sites/src/app/entities/landHistories.entity.ts
anton-bcgov: makes sense. done!

midhun-aot (approved):

midhun-aot (approved):

nikhila-aot (approved):

PR #35, opened 2024-08-13
SRS-442:Business logic to show banners to external users

nupurdixit13:
This PR is to include the logic to show the banner to external users who have purchased the site. Also few UI changes have been made to make sure the snapshot date appears as per design.

### DIFF
--- a/frontend/site-registry/src/app/features/details/SiteDetails.tsx
+++ b/frontend/site-registry/src/app/features/details/SiteDetails.tsx
@@ -557,16 +557,15 @@ const SiteDetails = () => {
           </div>
         )}
         <div className="section-details-header row">
-          <div className="d-flex">
-            <BannerDetails />
-            {UserType.External === userType &&
-              snapshot.status === RequestStatus.success &&
-              snapshot.snapshot.data !== null && (
-                <div className="py-2 snapshot">
-                  <span>{`Snapshot Taken: ${new Date(snapshot.snapshot.data[0].created)}`}</span>
-                </div>
-              )}
-          </div>
+          {UserType.External === userType &&
+            snapshot.status === RequestStatus.success &&
+            snapshot.snapshot.data !== null && (
+              <div>
+                <BannerDetails
+                  snapshotDate={`Snapshot Taken: ${new Date(snapshot.snapshot.data[0].created).toString().replace(/\s\([^)]+\)$/, '')}`}
+                />
+              </div>
+            )}
 
           {!isVisible && (
             <>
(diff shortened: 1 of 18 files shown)


frontend/site-registry/src/app/features/details/SiteDetails.tsx
```diff
@@ -557,16 +557,15 @@ const SiteDetails = () => {
...
-                  <span>{`Snapshot Taken: ${new Date(snapshot.snapshot.data[0].created)}`}</span>
-                </div>
-              )}
-          </div>
+          {UserType.External === userType &&
+            snapshot.status === RequestStatus.success &&
+            snapshot.snapshot.data !== null && (
+              <div>
+                <BannerDetails
+                  snapshotDate={`Snapshot Taken: ${new Date(snapshot.snapshot.data[0].created).toString().replace(/\s\([^)]+\)$/, '')}`}
```
midhun-aot: Please update this code. Only pass the required data from slice. Don't pass the complete snapshot slice. This code can cause error as there is no null check before referencing "created" property.  Also we are already using data-fns package to format the date. Please use formatDateRange / formatDate in utility.ts

midhun-aot (commented):
Please check the comments

frontend/site-registry/src/app/features/details/SiteDetails.tsx
nupurdixit13: I am only passing the created date as a prop to banner component, not sure if thats something wrong? also it has already been checked that the snapshot data shouldn't be null prior to that, however I can add `snapshot.snapshot.data?.length > 0` to avoid any potential errors...May be I am missing something so we can discuss that to have a clear understanding. 
Also the format of date does not match with the output of formatDate function present in utility.ts. I can add the function to just remove the timezone name at the end(which is required for this design) in the utility class and use however cannot use the one already present.

midhun-aot: 1. Use selectors to extract only the necessary parts of the state.  Here I could see snapshot slice is accessed 
   and then you are accessing snapshot.snapshot.data[0]. Instead of this approach you should only access snapshots array in Snapshots slice by using selectors.

2. I see the null check above. But there is no check to see array is empty or not before accessing the first element in array. Add check data.length > 0 before accessing first element.


3. Yes, If the current date helper functions are not matching the required formart. Please add a new method utility.ts so we can use later

frontend/site-registry/src/app/features/details/SiteDetails.tsx
```diff
@@ -587,9 +594,7 @@ const SiteDetails = () => {
           items={navItems}
           components={navComponents}
           dropdownItems={dropDownNavItems}
-          isDisable={false
-           
-          }
+          isDisable={false}
```
jaspalsingh-aot: isDisable = {
            UserType.External === userType && snapshot.snapshot.data === null
          }

Please add this condition as navigation pills will not be available everytime. it will be enable and disable based of snapshot is purchased or not.

nupurdixit13: Sure, added

midhun-aot (approved):

PR #36, opened 2024-08-15
Changes for  SRS 507

midhun-aot:


backend/sites/src/app/resolvers/snapshot/snapshot.resolver.ts
```diff
@@ -128,4 +128,50 @@ export class SnapshotsResolver {
...
+            return this.genericResponseProvider.createResponse(
+              `Failed to create snapshots. `,
+              422,
+              false,
+            );
+          }
+        }
+        else
+        {
+          return this.genericResponseProvider.createResponse(
```
nikhila-aot, line 163: this else is when the inputDTO is empty right? so have the error message appropriately

midhun-aot: done.

nikhila-aot (approved):

PR #45, opened 2024-08-27
feat/SRS-373 Implement parcel description GraphQL endpoint.

bolny:
# SRS-373 External - Parcel Description Page Purchased - Site Details

This update implements a graphql endpoint for parcel descriptions. The database query here is of particular note: since the view model (parcel description) is constructed from the data model (subdivision) in a complex manner, I did some performance profiling to determine that it is never worse to construct the parcel description on the database vs loading everything into node and doing it on the backend. 

I also refactored some of the existing response objects to be more reusable for my purposes. Let me know if this is appropriate.

Due to the size and complexity of this feature I'll be making a number of pull requests:

- [x] Database Migration
- [x] Backend functionality for internal users
- [ ] Shared component updates
- [ ] Front end functionality for internal users
- [ ] External user (snapshot) support for both front end and back end.


backend/sites/src/app/dto/response/response.dto.ts
```diff
@@ -1,9 +1,12 @@
-import { Type } from '@nestjs/common';
 import { Field, Int, ObjectType } from '@nestjs/graphql';
-import { GraphQLScalarType } from 'graphql';
+import { format } from 'date-fns';
 
 @ObjectType()
 export class ResponseDto {
+  constructor() {
+    this.timestamp = format(new Date(), 'MMMM do, yyyy');
+  }
```
nikhila-aot: should this be dd ?

backend/sites/src/app/resolvers/parcelDescription/parcelDescription.resolver.ts
```diff
@@ -0,0 +1,48 @@
...
+   * Find all parcel descriptions (subdivision entities) for a given site id.
+   * @param siteId site id
+   * @param page page number
+   * @param pageSize size of the page
+   * @param searchParam filtering term
+   * @param sortBy sorting term
+   * @param sortByDir sorting direction. Either ASC or DESC.
+   * @returns parcel descriptions (subdivisions) belonging to the given site.
+   */
+  @Roles({ roles: ['site-admin'], mode: RoleMatchingMode.ANY })
```
nikhila-aot: we will have to update the roles to the specific groups, namely External/Internal/Site Registrar. You can check notations resolver for reference

backend/sites/src/app/resolvers/parcelDescription/parcelDescription.resolver.ts
```diff
@@ -0,0 +1,48 @@
...
+  })
+  async getParcelDescriptionsBySiteId(
+    @Args('siteId', { type: () => Int }) siteId: number,
+    @Args('page', { type: () => Int }) page: number,
+    @Args('pageSize', { type: () => Int }) pageSize: number,
+    @Args('searchParam', { type: () => String }) searchParam: string,
+    @Args('sortBy', { type: () => String }) sortBy: string,
+    @Args('sortByDir', { type: () => String }) sortByDir: string,
+  ) {
+    return await this.parcelDescriptionService.getParcelDescriptionsBySiteId(
```
nikhila-aot: a good practice is to capture the return to a variable and handle errors on the return variable. Setting the return to a variable will be required when we incorporate logging into the code

nikhila-aot (changes requested):

backend/sites/src/app/dto/response/response.dto.ts
bolny: In this case `do` formats the day as "1st", "2nd", "3rd", "4th" and so on. I think I took that from somewhere else in the code, but when I think about it, I think the timestamp should be in ISO8601 format. Do you agree?

backend/sites/src/app/dto/response/response.dto.ts
```diff
@@ -1,9 +1,12 @@
-import { Type } from '@nestjs/common';
 import { Field, Int, ObjectType } from '@nestjs/graphql';
```
nikhila-aot: Also i just noticed that we have one more  dto  (baseHttpResponse) with the same fields, can we look at re-using it? or is there a reason for this new one ?

backend/sites/src/app/resolvers/parcelDescription/parcelDescription.resolver.ts
```diff
@@ -0,0 +1,68 @@
...
+        siteId,
+        page,
+        pageSize,
+        searchParam,
+        sortBy,
+        sortByDir,
+      );
+
+    let response = new ParcelDescriptionsResponse();
+    response.data = result.data;
```
nikhila-aot: any possibility of result  being null ?

backend/sites/src/app/resolvers/parcelDescription/parcelDescription.resolver.ts
```diff
@@ -0,0 +1,68 @@
...
+  })
+  async getParcelDescriptionsBySiteId(
+    @Args('siteId', { type: () => Int }) siteId: number,
+    @Args('page', { type: () => Int }) page: number,
+    @Args('pageSize', { type: () => Int }) pageSize: number,
+    @Args('searchParam', { type: () => String }) searchParam: string,
+    @Args('sortBy', { type: () => String }) sortBy: string,
+    @Args('sortByDir', { type: () => String }) sortByDir: string,
+  ) {
+    const result: ParcelDescriptionsServiceResult =
```
nikhila-aot: instead of using a new type, you can re-use the exisiting type genericResponseProvider (dto/response/genericResponseProvider.ts) An example of the use can be found in notation resolver.

nikhila-aot (changes requested):

nikhila-aot (dismissed):

nikhila-aot (approved):

PR #47, opened 2024-08-28
SRS-373 Implement getMostRecentSnapshot

bolny:
This PR implements a helper method on the snapshot service to retrieve the most recent snapshot by user id and site id. I intend to use this in the Parcel Description service when finding data for external users.

### DIFF
--- a/backend/sites/src/app/services/snapshot/snapshot.service.spec.ts
+++ b/backend/sites/src/app/services/snapshot/snapshot.service.spec.ts
@@ -534,6 +534,79 @@ describe('SnapshotService', () => {
     });
   });
 
+  describe('getMostRecentSnapshot', () => {
+    it('should return a single snapshot for a given userId and siteId', async () => {
+      const userId = '1';
+      const siteId = '1';
+      const res: Snapshots = {
+        id: 1,
+        userId: '1',
+        siteId: '1',
+        transactionId: '1',
+        whenCreated: new Date(2024, 8, 28),
+        whoCreated: 'ABC',
+        whenUpdated: new Date(2024, 8, 28),
+        whoUpdated: 'ABC',
+        site: sampleSites[0],
+        snapshotData: {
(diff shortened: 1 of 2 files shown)


backend/sites/src/app/services/snapshot/snapshot.service.spec.ts
```diff
@@ -534,6 +534,79 @@ describe('SnapshotService', () => {
...
+
+      await expect(
+        service.getMostRecentSnapshot(siteId, userId),
+      ).rejects.toThrow(errorMessage);
+      expect(snapshotRepository.findOne).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  describe('when there are no results', () => {
+    it('returns null', async () => {
```
bolny, line 592: This might be redundant but I think it's useful to assert that typeorm returning a null result will not cause an error in the snapshot service.

backend/sites/src/app/services/snapshot/snapshot.service.ts
```diff
@@ -72,6 +72,19 @@ export class SnapshotsService {
     }
   }
 
+  async getMostRecentSnapshot(siteId: string, userId: string) {
```
nupurdixit13, line 75: We already have a method (getSnapshotsBySiteId) to return the snapshots for siteId and userId in desc order. IMO you can use that and fetch the first record.

bolny: I agree that this is only a minor optimization, but I think it's useful to have around since it will be a common use case to fetch the most recent snapshot for the user/site. I think that it's more ergonomic to call `snapshot = snapshotService.getMostRecentSnapshot(siteId, userId)` than it is to call `snapshots = snapshotService.getSnapshotsBySiteId(siteId, userId); snapshot = snapshots.length > 0 ? snapshot[0] : null`.

nikhila-aot (approved):

PR #50, opened 2024-08-29
SRS-291: Export search results as .csv 

anton-bcgov:
In this PR:
1. Add the ability to export site search results as a csv file
[Sample File](https://github.com/user-attachments/files/16808034/export.3.csv)

2. Fix an issue that caused `SetNavList` to run infinitely. `getUser` returns a new object on every render, which is fine if you just want to read values from it. If, however, you try to introduce other side-effects based on the user object, it will be performed on every render potentially crashing the tab. [Link to commit](https://github.com/bcgov/nr-site-registry/commit/c51c9bef76bb05de0e77aafd221c6e96df168bc2)


midhun-aot (dismissed):
looks good

nikhila-aot (approved):

PR #69, opened 2024-09-24
Feat/srs 540

jaspalsingh-aot:


backend/sites/src/app/resolvers/dropdown/dropdown.resolver.ts
```diff
@@ -181,8 +183,8 @@ export class DropdownResolver {
       'DropdownResolver.getNotationParticipantRoleCd() start',
     );
     const result = await this.dropdownService.getNotationParticipantRoleCd();
-    if (result && result.length > 0) {
-      sitesLogger.info(
+    if (result.length > 0) {
```
nikhila-aot: (result && result.length > 0)

jaspalsingh-aot: done

PR #82, opened 2024-10-07
update getBannerType to work for internal users

bolny:
Restricting this endpoint to only external users causes me to get an error about an invalid banner type response whenever I try to view a site as an internal user. @nupurdixit13 I would appreciate it if you could let me know if this is an appropriate solution or if I should modify the front end to not make the request if the logged in user is internal.


bolny:
After chatting with @nupurdixit13 about this we decided it would be best to modify the front end code to not hit this endpoint in the case of an internal user.

PR #89, opened 2024-10-09
Refactor SiteDetails.saveSiteDetails to be Testable

bolny:
While working in this file I noticed that our test suite does not execute any code within the transaction block of `saveSiteDetails`. This is because the `entityManager` is mocked, so the inline callback is never executed. The existing test was giving a false positive result because mocks are always truthy.

I've extracted the core logic that is executed in the transaction block into a new function that can be tested directly. This will allow us to write tests for site saving.

### DIFF
--- a/backend/sites/src/app/services/site/site.service.ts
+++ b/backend/sites/src/app/services/site/site.service.ts
@@ -331,132 +331,17 @@ export class SiteService {
       if (!inputDTO) {
         return false;
       } else {
-        const {
-          siteId,
-          sitesSummary,
-          events,
-          eventsParticipants,
-          siteParticipants,
-          documents,
-          siteAssociations,
-          subDivisions,
-          landHistories,
-          profiles,
-        } = inputDTO;
-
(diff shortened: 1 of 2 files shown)


backend/sites/src/app/services/site/site.service.ts
```diff
@@ -434,6 +344,91 @@ export class SiteService {
...
+      sitesSummary,
+      events,
+      eventsParticipants,
+      siteParticipants,
+      siteAssociations,
+      subDivisions,
+      landHistories,
+      profiles,
+    } = inputDTO;
+
```
bolny, line 381: If I put a `return false;` here, the above spec will now fail. It did not previously.

anton-bcgov (dismissed):

midhun-aot:
Could you please resolve conflicts to merge it.

bolny:
@midhun-aot Done and Done.

midhun-aot (approved):

PR #92, opened 2024-10-16
SRS-361 - Implement parcel description handling for SaveSiteDetails

bolny:
This PR implements parcel description handling when saving a site. I made a couple of decisions here:

1. The front end does not have the concept of a `subdivision`, it only knows about parcel descriptions (parcel descriptions are a composite of the `siteSubdivision` and `subdivision` entities), so I made the decision for the API to accept the parcel description as it is handled by the front end. As a result I renamed the `subdivisionInputDTO` to `parcelDescriptionInputDTO` and modified it to fit the front end model.
3. I will write a subsequent PR that handles parsing the `parcelDescriptionInputDTO` into the relevant `subdivision` and `siteSubdivision` entities and committing them to the database. This is non-trivial.


backend/sites/src/app/services/site/site.service.ts
```diff
@@ -431,10 +432,14 @@ export class SiteService {
...
-      await transactionalEntityManager.save(Subdivisions, subDivisions);
+    if (parcelDescriptions) {
+      await this.parcelDescriptionService.saveParcelDescriptionsForSite(
+        inputDTO.siteId,
+        parcelDescriptions,
+        userInfo,
+      );
     } else {
-      console.log('No changes To Site subDivisions');
+      console.log('No changes To Parcel Descriptions.');
```
nikhila-aot: Please use the logger instead

nikhila-aot (changes requested):
Please resolve the conflicts in the PR

nikhila-aot (approved):

PR #93, opened 2024-10-16
chore(ci): fix Dockerfile naming conventions and updates

mishraomp:
First PR related to CI/CD, more to come around GitHub Actions and Helm Charts for OpenShift Deployment
1. Update of dockerfiles to align naming convention
2. Update of backend/site dockerfiles to move to latest LTS version of node and remove deprecated versions
3. Update of frontend/site-registry dockerfile to move to latest LTS version of node and nginx


backend/sites/Dockerfile
```diff
@@ -1,39 +1,27 @@
...
+
+# Copy over app
+WORKDIR /app
+COPY --from=build /app/node_modules ./node_modules
+COPY --from=build /app/dist ./dist
 
-# Expose port
-EXPOSE 4007
+# Port and health check
+EXPOSE 3000
```
bolny, line 23: I think this is going to conflict with the frontend port on our development devices. What would be a good way to mitigate this? (Or is there some docker networking magic that takes care of the issue?)

backend/sites/Dockerfile.migrations
```diff
@@ -21,5 +21,5 @@ COPY . .
 # Install postgresql-client
 RUN apk add --no-cache postgresql-client
 
-# Define the entrypoint
-ENTRYPOINT [ "sh", "initDB.sh" ]
\ No newline at end of file
+RUN chmod +x initDB.sh
+ENTRYPOINT ["/app/initDB.sh"]
```
bolny: I'm curious if this is just a convention or if there's something about executing shell scripts in a docker container that I don't know about which can cause issues.

bolny (dismissed):
I have a couple of questions but nothing that should be blocking.

backend/sites/Dockerfile
mishraomp: I see, the code is hardcoded to 4007, did not realise that, I will make it a env var, and then in local, you can change ports to avoid conflict

bolny: Let me double check that. I know the frontend currently runs on 3000, but I don't actually know where that number comes from.

bolny (approved):

PR #96, opened 2024-10-16
feat(CI): Added Github Actions code to build images and deploy helm charts.  

mishraomp:
currently it is set to PR mode for TESTING, , once tested , discuss with team, and modify workflows accordingly.

---

Thanks for the PR!

- [Backend](https://nr-site-registry-96-backendsites.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-96-frontendsiteregistry.apps.silver.devops.gov.bc.ca)

---

Thanks for the PR!

- [Backend](https://nr-site-registry-96-backend-sites.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-96-frontend-site-registry.apps.silver.devops.gov.bc.ca)


backend/sites/Dockerfile
```diff
@@ -18,10 +18,9 @@ ENV NODE_ENV=production
 WORKDIR /app
 COPY --from=build /app/node_modules ./node_modules
 COPY --from=build /app/dist ./dist
-
+COPY ./schema.graphql  ./schema.graphql
```
bolny: This file should be generated at runtime. In fact I just merged a PR to remove it from source control since it's constantly showing as dirty.

mishraomp: > This file should be generated at runtime. In fact I just merged a PR to remove it from source control since it's constantly showing as dirty.

Thanks a lot @bolny , you are a big saver.

backend/sites/src/main.ts
```diff
@@ -10,4 +11,8 @@ async function main() {
   });
   await app.listen(process.env.PORT || 4007);
 }
-main();
+main().then(()=>{
+  Logger.log('Application started successfully')
```
nikhila-aot: there is already a winston logger implemented in the application, we could replace with that as well

mishraomp: code updated to use winston logger.

.github/ISSUE_TEMPLATE/bug.md
acoard-aot, line 1: These issue templates look great. I'll bring them up next refinement to see if team wants to adjust any.

.github/ISSUE_TEMPLATE/epic.md
acoard-aot, line 1: We use Jira for our longer term tracking, I wonder if we will want to delete Epic. Though happy to keep it if team finds value, can discuss at refinement.

.github/codeowners
acoard-aot, line 1: We've discussed setting up Codeowners with more specific breakdowns in the future. This is a good start.

Out of curiosity, does this effectively result in no change? Or will this file as-is change current behaviour? Current behaviour is now we require a review, and the only valid people to make reviews are those that have edit access per`@bcgov/epd-team-ada` team. So I think it's unchanged (which is fine).

.github/workflows/.deployer.yml
```diff
@@ -0,0 +1,174 @@
...
+    outputs:
+      triggered:
+        description: 'Has a deployment has been triggered?'
+        value: ${{ jobs.deployer.outputs.triggered }}
+
+jobs:
+  deployer:
+    name: Helm
+    environment: ${{ inputs.environment }}
+    runs-on: ubuntu-24.04
```
acoard-aot, line 64: Minor point, but Ubuntu is now on 24.10. Perhaps could leave this for another task I do after accepting this PR, I won't consider this a blocking issue, just noting.

.github/workflows/.deployer.yml
```diff
@@ -0,0 +1,174 @@
...
+          # Helm release history
+          helm history ${{ steps.vars.outputs.release }}
+
+      ### Cleanup
+
+      # Completed pod cleanup
+      - if: steps.triggers.outputs.triggered == 'true'
+        run: |
+          # Completed pod cleanup
+          oc delete po --field-selector=status.phase==Succeeded || true
```
acoard-aot, line 174: Very minor quibble, I prefer `oc delete pod` as it's clearer to newer users. I know `po` is valid, but it saves one character and can be confusing to less experienced users. Happy to leave it too, just sharing perspective.

.github/workflows/.schema-spy.yml
```diff
@@ -0,0 +1,53 @@
...
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+      - name: Migrations
+        uses: docker://ghcr.io/bcgov/nr-site-registry/backend/sites:${{ inputs.tag }}
```
acoard-aot: I believe the migration image is called `sites-migration:dev`, at least in my original setup. However I think we could use the shared image with a different CMD

charts/app/templates/backend-sites/templates/deployment.yaml
```diff
@@ -0,0 +1,126 @@
...
+      annotations:
+        rollme: {{ randAlphaNum 5 | quote }}
+      labels:
+        {{- include "backendSites.labels" . | nindent 8 }}
+    spec:
+      {{- if .Values.backendSites.podSecurityContext }}
+      securityContext:
+        {{- toYaml .Values.backendSites.podSecurityContext | nindent 12 }}
+      {{- end }}
+      initContainers: # The container that will do the initial database seed and migrations.
```
acoard-aot, line 28: Great to see initContainer stuff here.

I'll note that this one IS using `sites-migration` as the image, whereas I think the earlier review comment spotted one just using `sites`. I assume these should be made the same?

charts/app/templates/_helpers.tpl
```diff
@@ -0,0 +1,47 @@
+{{/*
```
acoard-aot, line 1: Om, I would love you to walk me through this file a bit, and how it relates to the other _helpers.tpl files. Does this need to be invoked in a special way? Cheers.

charts/app/templates/backend-sites/templates/ingress.yaml
```diff
@@ -0,0 +1,27 @@
+{{- if .Values.backendSites.enabled }}
```
acoard-aot, line 1: Is there a reason we're using Ingress over Routes? K8's compatibility if going onto public cloud?

acoard-aot (dismissed):
Overall looks good, thanks Om.

I have a variety of questions in the review, but on the whole can approve it.

The only potential "bug"/mistake/etc I potentially identified (but may be wrong) is in `.github/workflwos/.schema-spy.yml`.  I'll note that file has a step called `migrations` that calls `backend/sites` rather than `backend/sites-migrations`. as it's image.  Elsewhere in the PR it seems to correctly call `backend/sites-migration`.  

I am still happy to approve the PR as-is and can review schemaspy later if it's affected.  I would ask you to please review my comments though, or maybe it's something we can sit and discuss together.

Thanks again Om, really appreciate you tackling this and sharing your expertise here.  Cheers.

.github/workflows/.schema-spy.yml
mishraomp: good catch, need to change it, I will do it along with the other change I am doing around frontend config, I just created the reusable action but currently it is not called from anywhere.

mishraomp:
> Overall looks good, thanks Om.
> 
> I have a variety of questions in the review, but on the whole can approve it.
> 
> The only potential "bug"/mistake/etc I potentially identified (but may be wrong) is in `.github/workflwos/.schema-spy.yml`. I'll note that file has a step called `migrations` that calls `backend/sites` rather than `backend/sites-migrations`. as it's image. Elsewhere in the PR it seems to correctly call `backend/sites-migration`.
> 
> I am still happy to approve the PR as-is and can review schemaspy later if it's affected. I would ask you to please review my comments though, or maybe it's something we can sit and discuss together.
> 
> Thanks again Om, really appreciate you tackling this and sharing your expertise here. Cheers.

Lets have a quick discussion on today's standup
1. when do we want to deploy to OS dev, test, prod ( ex: when merged to dev branch auto deploy to dev et..)
2. what are the things that should run on a PR (UT, Sonar) et..
3. Crunchy or Plain Postgres/postgis et..

charts/app/templates/backend-sites/templates/ingress.yaml
mishraomp: yes

charts/app/templates/_helpers.tpl
mishraomp: these files are auto invoked by HELM during template formation before deployment, so for each component it creates reusable labels and name and et..

charts/app/templates/backend-sites/templates/deployment.yaml
mishraomp: yeah the shcemaspy needs to be updated, this one is correct.

.github/workflows/.deployer.yml
mishraomp: yeah, right, we can take it up in another task after making sure it is available as a github runner.

.github/codeowners
mishraomp: codeowner makes it explicit to review for PR

.github/ISSUE_TEMPLATE/epic.md
mishraomp: sure , whatever needs to remove, we can, I will do that

acoard-aot:
Currently testing Keycloak configuration.  Have fixed our keycloak's redirect_uri, but I'm currently getting an Internal Server Error from the Siteminder url on login. Still investigating...

nikhila-aot (approved):

PR #97, opened 2024-10-16
Feat/srs 541 backend

jaspalsingh-aot:
**Overview**: This pull request introduces the backend API along with comprehensive test cases to ensure functionality and reliability.

**Changes Made**:

- Implemented the backend API for submitting the  data related to document.
- Added test cases to validate the new API endpoints and ensure expected behavior.
- Commented out certain properties in the entity definitions for SiteDoc and SiteDocPartics. These properties were removed to eliminate unnecessary relationships that are not required and have no visibility in the user interface.


backend/sites/schema.graphql
```diff
@@ -69,11 +69,28 @@ type SiteDocs {
   whoUpdated: String
   whenCreated: DateTime!
   whenUpdated: DateTime
-  rwmFlag: Float!
+  rwmFlag: Float
```
bolny: Suggestion: `schema.graphql` should not be checked into source control since it is an autogenerated file.

backend/sites/src/app/entities/siteDocPartics.entity.ts
```diff
@@ -24,15 +24,23 @@ export class SiteDocPartics extends ChangeAuditEntity {
   @Column('bigint', { primary: true, name: 'id' })
   id: string;
 
+  // @Field()
+  // @Column('character varying', { name: 'dpr_code', unique: true, length: 6 })
+  // dprCode: string;
```
bolny, line 29: Do we need to keep this code around for any reason? If not I think we should delete it. If we need to reference it in the future we can look at the git logs. The same applies to other areas I see commented code.

backend/sites/src/app/entities/siteDocPartics.entity.ts
```diff
@@ -24,15 +24,23 @@ export class SiteDocPartics extends ChangeAuditEntity {
...
+
+  //Need to remove unique constraint because as per stand-up discussion everyone
+  // will be ATH i.e Author. I would suggest to keep this null able as we don't have
+  // UI for dpr code.
```
bolny, line 33: I'm not sure that this information really needs to be kept in code. Maybe it would be more useful as a comment on the pull request or in a commit message. The same goes for the other entity files where I see a similar comment. What do you think?

backend/sites/src/app/services/site/site.service.ts
```diff
@@ -572,10 +563,10 @@ export class SiteService {
               ...siteDocumentParticipant,
               id: newDocParticId.toString(),
               sdocId: documentId,
-              rwmFlag: 0,
-              dprCode: dprCode ?? 'ATH',
+              // rwmFlag: 0,
+              dprCode: 'ATH', // dprCode is always ATH, we don't have UI and keeping this cloumn for maintaing historical data.
```
bolny: This is a good explanation for this hard coded value!

bolny:
```suggestion
              dprCode: 'ATH', // dprCode is always ATH. We don't have a UI for this value and keeping this column allows us to maintain historical data.
```

bolny (commented):
Everything here looks like it will work. Let me know what you think about the comments in the entity files and commented code.

backend/sites/schema.graphql
jaspalsingh-aot: Yeah, It suppose to be in git ignore but i am getting it in every merge.

backend/sites/src/app/entities/siteDocPartics.entity.ts
jaspalsingh-aot: I kept it because there may be sync issue. discussed with Nikhila, once we test the migration, we will remove the commented code.

backend/sites/schema.graphql
bolny: I'll take another look at it but, for now, can you checkout the copy from the dev branch so that it doesn't appear in this pull request please?

backend/sites/src/app/entities/siteDocPartics.entity.ts
jaspalsingh-aot: same for this, as dprcode was unique and a composite key, I make nullable because we don't need it as per current application UI. However keeping it commented, in case there will be sync issue, once everything get tested will remove the commented  code

backend/sites/schema.graphql
jaspalsingh-aot: done

bolny: Thank you!

bolny (approved):

PR #99, opened 2024-10-17
Add graphql codegen + Apollo Client

anton-bcgov:
This PR adds graphql codegen based on the schema exposed by the backend. In the frontend codebase, this allows for:

- Better developer experience:
When defining graphql queries/mutations, we can get suggestions about the fields available on the schema:
image

You also get compile-time warnings about missing or mistyped variables: 
image

### Important: 
**You need to have [Apollo GraphQL](https://marketplace.visualstudio.com/items?itemName=apollographql.vscode-apollo) extension installed in your code editor to get this experience**

- Improved type safety:
With every query/mutation defined in a `.graphql` file, the corresponding react hooks get generated for you.

For example, [the mapSearch query](https://github.com/bcgov/nr-site-registry/compare/anton/apollo-codegen?expand=1#diff-2dfc5cd5a3ee2809cfe5aa845824e363c1ea8f2d1b36e82b0505738bd98ac239), generates `useMapSearchQuery` and `useMapSearchLazyQuery`, where the return type of the data exactly matches the fields that are requested on the query and the types of those fields match what's defined on the backend schema. 

image


docker-compose.yml
```diff
@@ -75,7 +75,7 @@ services:
       retries: 3
       start_period: 40s
     hostname: frontend
-    image: node:14
+    image: node:16.14.0-alpine
```
bolny, line 78: I think @mishraomp was working on updating our docker images and we were updating to node 22. It might be a good idea to sync up with him. I assume that if this value gets updated later it won't hurt anything.

frontend/site-registry/src/graphql/generated.ts
bolny, line 1: Is this generated at runtime? I know that sometimes we have to have these generated files under source control but I really find it annoying. If it needs to be here that's fine too.

bolny (approved):
A couple of comments, but nothing that should block.

frontend/site-registry/src/graphql/generated.ts
anton-bcgov: We talked about this 1-1, posting here for posterity.

This file is generated in a separate background process (`graphql-codegen --watch`). The biggest reason why we're checking in this file into source control is that we need it for production builds. Otherwise, we'd need the graphql API already running when kicking off the front-end prod build. It's doable, but I suggest we move forward with a simpler approach for now.

PR #100, opened 2024-10-17
Changes For SRS-580

midhun-aot:
Changes for SRS-580 - Adding History For SR Approval

Updated column in History Log Table

Updated Roles For Resolver Methods getPendingSiteForSRApproval , bulkAproveRejectChanges


backend/sites/src/app/services/site/site.service.ts
```diff
@@ -1719,6 +1719,21 @@ export class SiteService {
       }
 
       this.sitesLogger.log('SiteService.processSRBulkUpdates() end');
```
nikhila-aot: i think this logger should be just before 'return true'

nikhila-aot (changes requested):

backend/sites/src/app/services/site/site.service.ts
midhun-aot: Done

nikhila-aot (approved):

PR #103, opened 2024-10-17
feat: SRS-552: Map Search - Map Icons

anton-bcgov:
In this PR:

- Load site data
- Display site markers based on the site's location. Each marker has three states: default, hover, and selected
- Add marker clustering

image

---

Thanks for the PR!

- [Backend](https://nr-site-registry-103-backend-sites.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-103-frontend-site-registry.apps.silver.devops.gov.bc.ca)

---

Thanks for the PR!

- [Backend](https://nr-site-registry-103-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-103-frontend.apps.silver.devops.gov.bc.ca)


frontend/site-registry/src/app/features/map/MapSearch.graphql
```diff
@@ -1,10 +1,10 @@
 query mapSearch($searchParam: String!) {
-    searchSites(searchParam: $searchParam, page: "1", pageSize: "100") {
+    searchSites(searchParam: $searchParam, page: "1", pageSize: "100000") {
```
anton-bcgov: For now, we're reusing `searchSites` query, but this will change in the PR to follow. I split FE and BE changes for simpler code review.

frontend/site-registry/src/app/features/map/SiteMarkers/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+    'site',
+    StringParam,
+  );
+
+  const moveToSiteLocation = useCallback(
+    (site: Site) => {
+      map.flyTo(
+        {
+          lat: site.latdeg || 0,
+          lng: site.longdeg ? site.longdeg * -1 : 0,
```
anton-bcgov: In the database, longitude values are positive, which is incorrect for locations in BC. I'll clarify if the prod data is the same way, and if so, we'll handle the conversion on the resolver level.

nikhila-aot: I can cross check in the oracle TEST db and update, but if I can recollect there were methods for conversion written in Oracle

nikhila-aot (commented):
Can we have the same folder naming convention (camel casing) for SiteMarkers?

frontend/site-registry/src/app/features/map/SiteMarkers/index.tsx
bolny: Ideally I think we should be handling this during our data import but I understand that things are going to get very complicated especially since it looks like we're going to need to do a two way sync. If there's no simple way to handle this at the database level then this is probably the best option since we know that latitude is always positive and longitude is always negative.

In any case I think a comment is warranted because I think that anyone unfamiliar with this code would probably be confused with this logic.

bolny: actually now that I'm thinking about it, it might be best to handle this on the backend when pulling the entities from the database if we can't do it on the database itself.

anton-bcgov:
> Can we have the same folder naming convention (camel casing) for SiteMarkers?

@nikhila-aot I renamed `SiteMarkers/index.tsx` to `siteMarkers/SiteMarkers.tsx`. Let me know if you had something else in mind.

frontend/site-registry/src/app/features/map/SiteMarkers/index.tsx
nikhila-aot: I did find few conversion logics written in oracle, and also a conversion script written by a previous resource that works on the migration scripts https://github.com/bcgov/nr-site-registry/blob/dev/ora2pg/modify_geo.py. 
@anton-bcgov we can go over it to see if does what you need.

frontend/site-registry/src/app/features/map/siteMarkers/icons.tsx
```diff
@@ -0,0 +1,22 @@
+import L from 'leaflet';
```
nupurdixit13: I would advise importing only the required class instead of global object.

frontend/src/app/features/map/siteMarkers/SiteMarker.tsx
```diff
@@ -0,0 +1,38 @@
+import { FC, memo, useState } from 'react';
+import { IconMarker } from '../IconMarker';
+import { LatLngLiteral } from 'leaflet';
+import { mapIconDefault, mapIconHover, mapIconSelected } from './icons';
```
nupurdixit13: Nit but I guess it will be useful: Can we rename it to mapMarkerIcon* since mapIcon is something relevant to map image which is being used in navs..Let me know your thoughts.

frontend/site-registry/src/app/features/map/SiteMarkers/index.tsx
nikhila-aot: @anton-bcgov  -  as discussed, we can keep the data in postgres db in the same format as what we receive now from oracle. This will also enable us to sync the data back to oracle seamlessly.

nupurdixit13 (approved):

PR #120, opened 2024-10-18
fix: SRS-361 implement autoincrementing ids for subdivision tables

bolny:
# Description

So the reason I wasn't able to insert new rows into the database for `subdivisions` and `site_subdivisions` I mentioned in standup is because TypeORM was inserting using `DEFAULT`. The problem is that when we imported the oracle db ddl, it didn't define any sort of auto incrementing ids for these tables.

This PR fixes this problem by:
1. using TypeORM's `@PrimaryGeneratedColumn` decorator on the relevant primary key columns so that TypeORM will not attempt to set the id of a new row.
2. migrating the database to create new sequences for the tables primary keys, altering those tables to use the new sequences, and setting the sequences current values to the maximum value that is already in the database (to avoid breaking existing databases).

Fixes # SRS-361 (partially)

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Migrated my exisiting database and verified that I was able to insert a new row with the default id value in both tables.
- [x] Migrated a fresh database and verified that I was able to insert a new row with the default id value in both tables.
- [x] Inserted new rows into the database to ensure that the ids properly increment.

## Checklist

- [x] I have read the [CONTRIBUTING](CONTRIBUTING.md) doc
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have already been accepted and merged

Our migrations and decorators are not testable so I have not added any new tests.

## Further comments

@nikhila-aot I'm not sure how this change will impact our oracle data import. We may need to run the following code on the postgresql database to set the sequences correctly after import:

```
SELECT setval('sites.subdivision_id_seq', (SELECT MAX(id) FROM "sites"."subdivisions"));
SELECT setval('sites.site_subdivision_site_subdiv_id_seq', (SELECT MAX(site_subdiv_id) FROM "sites"."site_subdivisions"));
```

Alternatively: if this will impact data import, I can write a sequence generator in the backend to handle setting ids in code rather than on the database. Please let me know.

---

Thanks for the PR!

- [Backend](https://nr-site-registry-120-backend-sites.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-120-frontend-site-registry.apps.silver.devops.gov.bc.ca)

---

Thanks for the PR!

- [Backend](https://nr-site-registry-120-backend-sites.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-120-frontend-site-registry.apps.silver.devops.gov.bc.ca)

---

Thanks for the PR!

- [Backend](https://nr-site-registry-120-backend-sites.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-120-frontend-site-registry.apps.silver.devops.gov.bc.ca)

### DIFF
--- a/backend/sites/src/app/entities/siteSubdivisions.entity.ts
+++ b/backend/sites/src/app/entities/siteSubdivisions.entity.ts
@@ -1,5 +1,12 @@
 import { Field, ObjectType } from '@nestjs/graphql';
-import { Column, Entity, Index, JoinColumn, ManyToOne } from 'typeorm';
+import {
+  Column,
+  Entity,
+  Index,
+  JoinColumn,
+  ManyToOne,
+  PrimaryGeneratedColumn,
+} from 'typeorm';
 import { Sites } from './sites.entity';
 import { Subdivisions } from './subdivisions.entity';
 import { ChangeAuditEntity } from './changeAuditEntity';
@@ -63,7 +70,7 @@ export class SiteSubdivisions extends ChangeAuditEntity {
   sprofDateCompleted: Date | null;
 
   @Field()
-  @Column('bigint', { primary: true, name: 'site_subdiv_id' })
+  @PrimaryGeneratedColumn('increment')
   siteSubdivId: string;
 
   @Field()
--- a/backend/sites/src/app/entities/subdivisions.entity.ts
+++ b/backend/sites/src/app/entities/subdivisions.entity.ts
@@ -1,5 +1,11 @@
 import { Field, ObjectType } from '@nestjs/graphql';
-import { Column, Entity, Index, OneToMany } from 'typeorm';
+import {
+  Column,
+  Entity,
+  Index,
+  OneToMany,
+  PrimaryGeneratedColumn,
+} from 'typeorm';
 import { SiteSubdivisions } from './siteSubdivisions.entity';
 import { ChangeAuditEntity } from './changeAuditEntity';
 
@@ -9,7 +15,7 @@ import { ChangeAuditEntity } from './changeAuditEntity';
 @Entity('subdivisions')
 export class Subdivisions extends ChangeAuditEntity {
   @Field()
-  @Column('bigint', { primary: true, name: 'id' })
+  @PrimaryGeneratedColumn('increment')
   id: string;
 
   @Field()
--- a/backend/sites/src/migrations/1729275924909-master-script.ts
+++ b/backend/sites/src/migrations/1729275924909-master-script.ts
@@ -1,0 +1,39 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class MasterScript1729275924909 implements MigrationInterface {
+  name = 'MasterScript1729275924909';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE SEQUENCE "sites"."subdivision_id_seq" OWNED BY "sites"."subdivisions"."id"`,
+    );
+    await queryRunner.query(
+      `SELECT setval('sites.subdivision_id_seq', (SELECT MAX(id) FROM "sites"."subdivisions"))`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "sites"."subdivisions" ALTER COLUMN "id" SET DEFAULT nextval('sites.subdivision_id_seq')`,
+    );
+    await queryRunner.query(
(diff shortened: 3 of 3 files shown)


backend/sites/src/app/entities/siteSubdivisions.entity.ts
```diff
@@ -63,7 +70,7 @@ export class SiteSubdivisions extends ChangeAuditEntity {
   sprofDateCompleted: Date | null;
 
   @Field()
-  @Column('bigint', { primary: true, name: 'site_subdiv_id' })
+  @PrimaryGeneratedColumn('increment')
```
jaspalsingh-aot, line 73: is this a new primary column? if not, this might create sync issue if siteSubdivId was not set as auto generated. Besides, will the column name be 'site_subdiv_id' ?

backend/sites/src/app/entities/subdivisions.entity.ts
```diff
@@ -9,7 +15,7 @@ import { ChangeAuditEntity } from './changeAuditEntity';
 @Entity('subdivisions')
 export class Subdivisions extends ChangeAuditEntity {
   @Field()
-  @Column('bigint', { primary: true, name: 'id' })
+  @PrimaryGeneratedColumn('increment')
```
jaspalsingh-aot, line 18: is this a new primary column? if not, this might create sync issue if id was not set as auto generated.

backend/sites/src/app/entities/siteSubdivisions.entity.ts
bolny: These are both the existing primary columns for these tables. All this line does is tell TypeORM how to build its queries. In this case:

before: `insert into sites.sitesubdivisions (site_subdiv_id, ...<everything else>) values (DEFAULT, ...<everything else>);`

after: `insert into sites.sitesubdivisions (...<everything else>) values (...<everything else>);`

This is because that particular decorator tells typeORM to let the database handle it.

backend/sites/src/app/entities/subdivisions.entity.ts
bolny: These are both the existing primary columns for these tables. See my above explanation about what this line does.

backend/sites/src/app/entities/siteSubdivisions.entity.ts
nikhila-aot: The only doubt I have here is with respect to data migration from oracle to postgres. Since oracle already have values for this column, and we make this column auto generated in postgres, whether this would cause an error for the migration

nikhila-aot: Also you might have to add the property  name: 'site_subdiv_id'  to the PrimaryGeneratedColumn

bolny: I'll double check the column name because I didn't test updating with this, only adding. Thanks for calling that out.

Regarding the data import: I'm concerned about it as well. I don't know about oracle databases, so I don't know how it assigns primary key values. In postgres you typically use a sequence generator like I've defined here. These sequences aren't in the database dump that you provided me. If the oracle database uses sequences like postgres, and ora2pg will do a two way sync of the sequences then this isn't actually an issue. If that's the case then I can remove this migration and just update the entity definitions so that they don't try to set the id to a specific value with the understanding that a production database will handle things correctly.

Is there any way for us to get access to the oracle db to check and also verify the behaviour of ora2pg?

bolny: Actually, I think the best thing to do here is to wait until the oracle db import is complete, dump the production db and use that to build off of. I'll just continue doing speculative work until that is complete.

bolny:
After consideration I'm going to close this PR until we have a functional production database that I can base my work off of.

backend/sites/src/app/entities/siteSubdivisions.entity.ts
mishraomp: @bolny , you can raise a SD ticket with approval from Kerri to have readonly access to your IDIR to prod database. Let me know if you need help.

PR #145, opened 2024-10-30
fix: SRS-617 - Prevent showing banner related error message for internal users

nupurdixit13:
This fix is to make sure we don't get any error when trying to fetch the bannerType. Currently the error was coming due to two reasons:
1. The default state for userType was set to "External" thus the two useEffects using the userType were picking two different values intermittently.
2. The dependency array for the useEffects was not using the loggedInuser and userType which is necessary to maintain consistent values throughout. 

Key Changes:
1. Dependency Array in First useEffect: Added loggedInUser to the dependency array to ensure it re-runs when loggedInUser changes.
2. Dependency Array in Second useEffect: Added userType to the dependency array to ensure it re-runs when userType changes.

---

Thanks for the PR!

- [Backend](https://nr-site-registry-145-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-145-frontend.apps.silver.devops.gov.bc.ca)

### DIFF
--- a/frontend/src/app/features/details/SiteDetails.tsx
+++ b/frontend/src/app/features/details/SiteDetails.tsx
@@ -166,7 +166,7 @@ const SiteDetails = () => {
   const [showLocationDetails, SetShowLocationDetails] = useState(false);
   const [showParcelDetails, SetShowParcelDetails] = useState(false);
   const [save, setSave] = useState(false);
-  const [userType, setUserType] = useState<UserType>(UserType.External);
+  const [userType, setUserType] = useState<UserType | null>(null);
   const [viewMode, setViewMode] = useState(SiteDetailsMode.ViewOnlyMode);
   const [isLoading, setIsLoading] = useState(true);
   const dispatch = useDispatch<AppDispatch>();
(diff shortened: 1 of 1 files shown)


mishraomp:
@nupurdixit13 , could we please use conventional commit pattern for PR title
https://www.conventionalcommits.org/en/v1.0.0/#summary

nupurdixit13:
> @nupurdixit13 , could we please use conventional commit pattern for PR title https://www.conventionalcommits.org/en/v1.0.0/#summary

@mishraomp : I remember this being discussed in the past but wasn't sure if we have already started implementing it. Thanks for reminding! Let me know if the new message adheres to the standards.

mishraomp:
> > @nupurdixit13 , could we please use conventional commit pattern for PR title https://www.conventionalcommits.org/en/v1.0.0/#summary
> 
> @mishraomp : I remember this being discussed in the past but wasn't sure if we have already started implementing it. Thanks for reminding! Let me know if the new message adheres to the standards.

thanks @nupurdixit13 , yeah, this works great

nikhila-aot (approved):

PR #186, opened 2024-11-07
feat: SRS-527 Map Search – Find Me

nupurdixit13:
As a user, I want to use Find Me so that I can locate my location on the map.

Description:
When a user selects Find Me, a browser notification will appear with text “[<URL>.gov.bc.ca] wants to know your location” a user is able to block or allow selection. 

When a user follows through with these actions, the map shows where a user is.

Tasks remaining: Unit tests and few changes in css(hover/onClick based on designs being prepared by Harry)

---

Thanks for the PR!

- [Backend](https://nr-site-registry-186-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-186-frontend.apps.silver.devops.gov.bc.ca)


frontend/src/app/features/map/map-slice.ts
anton-bcgov, line 1: I think that hooking into redux for a single boolean field is a massive overkill and a regular state field on the level of `<MapView />` component should do the trick just fine. 

You mentioned that you were getting map context errors before the redux implementation, but looking at the code just now, I'm 99% certain those were unrelated. Happy to hop on a call to debug this together

frontend/src/app/features/map/FindMeControl.tsx
```diff
@@ -0,0 +1,35 @@
+import { useDispatch } from 'react-redux';
+import { IconButton } from '@mui/material';
+import clsx from 'clsx';
+
+import { setMyLocationVisible, useMyLocationVisible } from './map-slice';
+import { useGeolocationPermission } from '../../../hooks/useMyLocation';
+
+import { FindMe } from '../../components/common/icon';
+
+export function FindMeControl() {
```
anton-bcgov, line 10: `FindMeControl` and `FindMeButton` are almost identical. To avoid code duplication, they should be combined into a single component that accepts settings for its appearance/position as props.

anton-bcgov (changes requested):

frontend/src/app/features/map/FindMeControl.tsx
nupurdixit13: The two are only similar to check for position but are very much different in terms of returning the component altogether. Its just not only the difference in styling which could be passed as props. I can definitely create a common component but it will a lot of overhead to pass numerous props and the added advantage is only these 2 components. Not sure if that would be a very effective approach. We can discuss more offline.

PR #189, opened 2024-11-10
fix: migrations

### LINKED ISSUE #126: Make the initDB more stable and compiled.
currently the initDB is not stable and is compiling using ts-node during run time.

Crunchy DB takes around 5 to 6 minutes to start, the initDB migration script currently fails to support the delay start.

I manually removed the backend pods and then restarted when crunchy was up.

need to build some smartness into migration.

mishraomp:
1. add retry with backoff
2. run compiled migrations

---

Thanks for the PR!

- [Backend](https://nr-site-registry-189-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-189-frontend.apps.silver.devops.gov.bc.ca)

### DIFF
--- a/backend/Dockerfile.migrations
+++ b/backend/Dockerfile.migrations
@@ -14,11 +14,6 @@ RUN npm ci
 
 # Bundle app source
 COPY . .
-#build to make the compilation happens to JS
-RUN npm run build
-
-# move src folder from dist to src and overwrite the existing src folder
-RUN mv dist/src src
 # Install postgresql-client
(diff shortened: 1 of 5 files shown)


mishraomp:
currently tsnode is required and typeorm does not support just js migrations, which is sad and unfortunate as it adds runtime overhead and latency. will revert back the code to use TS.
https://github.com/typeorm/typeorm/issues/5535

nikhila-aot (approved):

PR #192, opened 2024-11-13
fix: Change to Display Folio Search Functionality in Mobile View

midhun-aot:
- Shows action dropdown for mobile view ( with add to cart and add to folio button)
- 

---

Thanks for the PR!

- [Backend](https://nr-site-registry-192-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-192-frontend.apps.silver.devops.gov.bc.ca)


frontend/src/app/components/search/SearchInput.tsx
```diff
@@ -101,6 +107,17 @@ const SearchInput: React.FC<ISearchInput> = ({
                   </div>
                 );
               })}
+              </div>
+                {showCloseBtnInDropdownOptions && <div
+                    key={10000}
```
jaspalsingh-aot: do we need key as hard coded ?

frontend/src/app/features/details/SiteDetails.tsx
```diff
@@ -497,6 +499,49 @@ const SiteDetails = () => {
   if (snapshot.status === RequestStatus.failed)
     return <div>Error: {snapshot.error || 'Failed to load data'}</div>;
 
+
+  const renderOptionsForExternalUser = () => {
+    if(viewMode === SiteDetailsMode.ViewOnlyMode &&
+      userType === UserType.External)
+      { return ( <> <div className='d-block d-sm-none'>
```
anton-bcgov: please run `npm run format` on your changes

frontend/src/app/features/folios/AddToFolio.tsx
```diff
@@ -26,14 +28,24 @@ const AddToFolio: FC<AddToFolioProps> = ({
...
   const auth = useAuth();
 
   const [getFolioItems, { loading, data }] =
     useFolio_GetFolioItemsForUserLazyQuery();
 
+   useEffect(()=>{
+    if (showSearchBoxOnly) {
+      getFolioItems();
+    }
+   }, [])
```
anton-bcgov: This `useEffect` should have dependencies:

```suggestion
   useEffect(() => {
    if (showSearchBoxOnly) {
      getFolioItems();
    }
  }, [getFolioItems, showSearchBoxOnly]);
```

frontend/src/app/features/folios/AddToFolio.tsx
```diff
@@ -77,51 +117,52 @@ const AddToFolio: FC<AddToFolioProps> = ({
...
+  return (
+    <div className="d-flex  flex-column justify-content-center">
+        {renderSearchInput(true)}
+      <div
+        onClick={handleClose}
+        className="d-flex flex-row align-items-center pt-2 justify-content-center "      >        
+        <XmarkIcon className="custom-search-label"></XmarkIcon>
+        <span className='custom-search-label'>Close</span>
+      </div>
+    </div>
```
anton-bcgov: So for this to look right on the page, the `AddToFolio` component has to have `ModalDialogWrapperWithHeader` as its parent, and I'm not a fan of this approach.

What I think we should do instead is to render the modal in the `overlay` props based on the screen size. The idea here is something like

```
const isSmall = useMediaQuery('@media (max-width:899.95px)'); // not actual code, please don't use this

return (
  <OverlayTrigger
      overlay={isSmall ? _ search input within modal here_ : _ regular popup _}
  />
)
```

This should also take care of my other concern: your PR changes the behavior of one "Add to Folio" button, but we have at least two more of them in the app (in Summary and Search). Placing the logic in the Add To Folio component itself saves us the trouble of modifying other components and adding more state variables in them, making it more maintainable in the long run.

anton-bcgov (changes requested):

jaspalsingh-aot (changes requested):

frontend/src/app/features/folios/AddToFolio.tsx
midhun-aot: removed as no longer needed with updated code

frontend/src/app/components/search/SearchInput.tsx
midhun-aot: yes, its an close button added to the bottom of the list of options

frontend/src/app/features/folios/AddToFolio.tsx
midhun-aot: Thanks for the feedback and the pair programming we did on this component.

anton-bcgov (dismissed):

jaspalsingh-aot (dismissed):
code seems fine, however GitHub actions frontend unit test is not passed

jaspalsingh-aot (approved):
seems fine to me

PR #193, opened 2024-11-13
Fixed larger device side bar hover UI issue

midhun-aot:
- Fixed CSS Issue on larger device when mouse hover on sidebar


bolny (approved):
Looks good.

PR #207, opened 2024-11-18
fix: SRS-624 Fix Parcel Descriptions API issue

bolny:
# Description

This fix does three things:

- Refactors the Parcel Descriptions component to be more in line with react/redux best practices by updating a state object in response to user input and api results. 
- Adds a flag to the state to determine whether or not an API call is needed.
- Updates the date noted column to use our datetime component, removing the need for my custom date parsing code. (something Midhun pointed out in in times immemorial)

Fixes # SRS-624

## Type of change

Please delete options that are not relevant.

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Manual toggling of all inputs to ensure proper behaviour.
- [x] Browsing between multiple sites to ensure that the correct data for the current site is displayed.
- [x] Unit testing.

## Checklist

- [x] I have read the [CONTRIBUTING](CONTRIBUTING.md) doc
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have already been accepted and merged

## Further comments

---

Thanks for the PR!

- [Backend](https://nr-site-registry-207-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-207-frontend.apps.silver.devops.gov.bc.ca)


frontend/src/app/features/details/parcelDescriptions/parcelDescriptionsSlice.ts
```diff
@@ -1,163 +1,102 @@
...
+          searchParam: currentState.searchParam,
+          sortBy: currentState.sortBy,
+          sortByDir: currentState.sortByDir,
+          pending: false, // This is only used for SR approval in the site details component.
         },
       });
     } catch (error) {
       throw error;
     }
     if (response?.status != 200) {
```
midhun-aot: I think, if you receive any other response code other than 200 for the request, fetchParcelDescriptions.rejected reducer will be called. So I think this is not needed here. But if you are checking for the httpStatusCode which is returned from the API then its better to check and set the status in fetchParcelDescriptions.fulfilled

frontend/src/app/features/details/parcelDescriptions/parcelDescriptionsSlice.ts
```diff
@@ -1,163 +1,102 @@
...
-          id: parcelDescription?.id,
-          descriptionType: parcelDescription?.descriptionType,
-          idPinNumber: parcelDescription?.idPinNumber,
-          dateNoted: formattedDateNoted,
-          landDescription: parcelDescription?.landDescription,
-        };
-      },
-    );
+    const responseData = response.data?.data
+      ?.getParcelDescriptionsBySiteId as IParcelDescriptionResponseDto;
```
midhun-aot: This code will work here, but in my opinion I would move this logic to fetchParcelDescriptions.fulfilled reducer.

frontend/src/app/features/details/parcelDescriptions/parcelDescriptions.tsx
```diff
@@ -1,59 +1,74 @@
...
+    data,
+    requestStatus,
+    totalResults,
+    currentPage,
+    resultsPerPage,
+    searchParam,
+    sortBy,
+    sortByDir,
+    sortByInputValue,
+    needsUpdate,
```
midhun-aot: Every time the Redux state updates, useSelector will re-run its selector function. So here as you are selecting the root state , IMO every time any state changes in application will cause parcel description component to re-render, which is not ideal and lead to performance and other issues. I would recommend to move this code to parcel description slice and expose Individual selectors as needed. (Similar way we implemented in other components)

frontend/src/app/features/details/parcelDescriptions/parcelDescriptions.tsx
```diff
@@ -1,59 +1,74 @@
...
   const handleSelectPage = (newPage: number) => {
-    dispatch(setCurrentPage(newPage));
+    currentPage = newPage;
+    needsUpdate = true;
+    fetchNewParcelDescriptions();
   };
 
   const handleChangeResultsPerPage = (newResultsPerPage: number) => {
-    dispatch(setResultsPerPage(newResultsPerPage));
+    resultsPerPage = newResultsPerPage;
```
midhun-aot: we should manage and track state changes react way using useState() otherwise will lead to bugs and unpredictable behaviors.

frontend/src/app/features/details/parcelDescriptions/parcelDescriptionsSlice.ts
```diff
@@ -1,163 +1,102 @@
...
-  sortBy: defaultValues.sortBy,
-  sortByDir: defaultValues.sortByDir,
-  sortByInputValue: defaultValues.sortByInputValue,
+  needsUpdate: true,
 };
 
 export const fetchParcelDescriptions = createAsyncThunk(
   'parcelDescriptions/fetchParcelDescriptions',
-  async (params: IFetchParcelDescriptionParams) => {
+  async (currentState: IParcelDescriptionsState) => {
```
midhun-aot: IMO, If the other comments are fixed and state is managed in React way. We don't need this logic here.

frontend/src/app/features/details/parcelDescriptions/parcelDescriptions.tsx
```diff
@@ -64,61 +79,74 @@ const ParcelDescriptions = () => {
...
   };
 
-  useEffect(() => {
-    const params: IFetchParcelDescriptionParams = {
-      siteId: siteId,
-      page: currentPage,
-      pageSize: resultsPerPage,
+  const fetchNewParcelDescriptions = () => {
+    if (needsUpdate) {
+      requestStatus = RequestStatus.loading;
```
midhun-aot: Similar to my above comment, this should be maintained in slice and only modified in an reducer so it should be dispatched as an action.

midhun-aot (changes requested):

midhun-aot (approved):

PR #217, opened 2024-11-22
fix: SRS-672: transform site's `longdeg` values

anton-bcgov:
# Description

This change centralizes the logic of converting positive `longdeg` coordinate values into negative.

Fixes [SRS-672](https://env-epd.atlassian.net/browse/SRS-556)

## Type of change

# How Has This Been Tested?

- [x] Navigate to Map Search page and verify that site markers appear as before and "zoom to" pans the map to the correct place.

---

Thanks for the PR!

- [Backend](https://nr-site-registry-217-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-217-frontend.apps.silver.devops.gov.bc.ca)


backend/src/app/entities/sites.entity.ts
```diff
@@ -109,7 +109,16 @@ export class Sites extends ChangeAuditEntity {
...
 
   @Field({ nullable: true })
-  @Column('double precision', { name: 'longdeg', nullable: true })
+  @Column('double precision', {
+    name: 'longdeg',
+    nullable: true,
+    transformer: {
+      // Convert positive longdeg values into negative (and therefore valid BC coordinates)
+      // In this context, `from` runs when reading from the database.
+      from: (value) => (value !== null && value > 0 ? -value : value),
```
nikhila-aot, line 118: will this work for the existing data in the db? also if at all we have to sync back to oracle, we maybe have to write the reverse transformation

anton-bcgov: >will this work for the existing data in the db?

yep

>also if at all we have to sync back to oracle, we maybe have to write the reverse transformation

If the sync will happen through the application layer and hit the Site entity we have declared here, then yes, we'll need to transform it back. Otherwise, no.

nupurdixit13 (approved):

PR #221, opened 2024-11-28
fix: edit button view for small screen when click edit/sr mode

jaspalsingh-aot:
# Description
Fix action button for mobile screen.

---

Thanks for the PR!

- [Backend](https://nr-site-registry-221-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-221-frontend.apps.silver.devops.gov.bc.ca)


anton-bcgov (changes requested):
The layout change (the save/cancel button appearing in the dropdown) looks good to me.

I disagree with this change, however:

https://github.com/user-attachments/assets/e224fdf4-e98b-4dac-b459-e48eafae8b0c

Buttons should not appear "disabled" until they are hovered on. I'm 100% sure that the intention in the designs was that the button should be disabled until there are actual changes that can be saved.

jaspalsingh-aot:
> The layout change (the save/cancel button appearing in the dropdown) looks good to me.
> 
> I disagree with this change, however:
> 
>  Screen.Recording.2024-11-28.at.12.03.22.PM.mov 
> Buttons should not appear "disabled" until they are hovered on. I'm 100% sure that the intention in the designs was that the button should be disabled until there are actual changes that can be saved.

The scope of this ticket was primarily focused on ensuring the layout matches the design, which is why the save/cancel buttons were implemented as per the current view. I understand your point about the button's disabled state, and I agree that it's important for the button to remain inactive until there are changes to save.

I haven't seen any active state of the button in the design, but I can create a new ticket to address the functionality aspect of the buttons, where we can implement this behavior as per the design intent. Let me know if that works for you!

PR #224, opened 2024-11-28
feat: SRS-556: Changes to support text search results drawer

anton-bcgov:
# Description

In this PR:

- Move common map pan and zoom settings into a separate file
- Centralize the way we read and set URL query params in the map search view - `MapSearchQueryParamsContext`. This removes the need to re-declare query params getters and setters in each component that uses them and lets us better manage setting and clearing params conditionally. 
  - For example: when a search query is applied, user should be able to select a site on the map, which updates the query to this: `map?search=test&site=7112` - site is selected, search query is kept in place. However, when updating the search query, the siteId should be removed from the URL.
- Add `MapSearchDrawer` that manages open/close state of the drawer and which content is displayed in the drawer based on the query params. Originally, I planned to have `SiteDetailsDrawer` and `SearchResultsDrawer` to have their own `<Drawer>` components, but managing two (and potentially more) drawers that can overlay one another turned out to be a big pain. 

**_NOT_** in this PR:

- City / Site autocomplete dropdown options. This will be tackled in a separate ticket
- Displaying the actual search result data. This will be tackled in the upcoming PR, for now, the drawer displays raw JSON.

Demo:

https://github.com/user-attachments/assets/9268ea5c-b66c-4f67-b321-3297072c4787

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [x] Type in the search query and press enter. If valid results are found, the map should pan and zoom to the bounds containing the site markers and the search results drawer should appear
- [x] With a search query present, selecting a marker should display the site details drawer
- [x] Updating the search query should unset the `site` query param
- [x] Refreshing the page with `search` query in the URL should apply the same search term to the GraphQL query and navigate the map accordingly

---

Thanks for the PR!

- [Backend](https://nr-site-registry-224-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-224-frontend.apps.silver.devops.gov.bc.ca)


frontend/src/app/features/map/siteDrawer/SiteDetailsDrawerContent.tsx
```diff
@@ -0,0 +1,157 @@
...
+  if (d) result += `${d}d`;
+  if (m) result += `, ${m}m`;
+  if (s) result += `, ${s}s`;
+  return result;
+};
+
+interface SiteDetailsDrawerContentProps {
+  mapRef: RefObject<Map | null>;
+}
+export const SiteDetailsDrawerContent: FC<SiteDetailsDrawerContentProps> = ({
```
anton-bcgov, line 62: This component is mostly unchanged compared to its previous iteration - `SiteDetailsDrawer`. The _only_ difference is that it doesn't render its own `<Drawer>` component anymore.

jaspalsingh-aot (commented):
Do we need test cases here or  that will be a separate PR

anton-bcgov:
> Do we need test cases here or that will be a separate PR

Tests will be added in a separate PR

jaspalsingh-aot (dismissed):
seems fine too me

nikhila-aot (approved):

PR #234, opened 2024-12-12
feat: SRS-556: Sites table in the search results drawer

anton-bcgov:
# Description

In this PR:

- Add sites table and column / filters tabs in SearchResultsDrawerContent component
- Add `siteIds` arg to `searchSites` query. The intended flow here:
  - Text search triggers `mapSearch` query that returns site ids and the geolocation data for the map markers
  - In the drawer, `searchSites` then runs and returns detailed data about the site ids from step 1 to be displayed in the table. 
  - Additional filters can then be applied to sites from step 1.
- Change `searchSites` pagination-related args' types to Int instead of String
- Break down some of the parts of the `/search` page into components that can be reused elsewhere: `SearchResultsFilters` `SearchResultsActions`

_NOT_ in this PR:
- Pagination of the table in the drawer
- Filters - the fields are displayed, but the filters themselves are not used in the query yet
- FilterPills component hasn't been ported from the search page yet

## Demo

Screenshot 2024-12-11 at 4 47 22 PM

Screenshot 2024-12-11 at 4 47 36 PM

---

Thanks for the PR!

- [Backend](https://nr-site-registry-234-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-234-frontend.apps.silver.devops.gov.bc.ca)


bolny (approved):
Looks good to me and I learned a thing or two about screen querying for specs!

PR #235, opened 2024-12-13
feat: SRS-556: Map search results table pagination + `icon` button variant

anton-bcgov:
# Description

In this PR: 
- Add pagination to map search result table
- Add 'icon' button variant that matches the design system ([Figma link](https://www.figma.com/design/ds1JjB8vYykPb6SF1RskfG/EPD-Design-System---Bespoke?node-id=574-1406&t=mO8Q8cETEmToHYlv-1))
- Make Table's pagination controls actual buttons instead of clickable divs - **_this affects all tables with pagination!_**

Demo:

https://github.com/user-attachments/assets/13a66190-ebf6-4483-9ebc-6cd7e3947fa9

---

Thanks for the PR!

Deployments, as required, will be available below:
- [Backend](https://nr-site-registry-235-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-235-frontend.apps.silver.devops.gov.bc.ca)

Please create PRs in draft mode.  Mark as ready to enable:
- [Analysis Workflow](https://github.com/bcgov/nr-site-registry/actions/workflows/analysis.yml)

After merge, new images are deployed in:
- [Merge Workflow](https://github.com/bcgov/nr-site-registry/actions/workflows/merge.yml)

PR #240, opened 2024-12-18
feat: SRS-556: Map search results table filters

anton-bcgov:
# Description

This PR is (hopefully) the final part of the text search + search results table work and includes:
- Search results can now be filtered - applying filters to the table should not affect the sites returned by the mapSearch query
- Refactor the `searchSites` query - instead of having 19 optional arguments we now have 1 object containing those optional arguments. This step also fixes type inconsistencies between the exposed argument types and actual types on the backend. For example, string vs number types in this definition `@Args('latDegrees', { type: () => String, nullable: true }) latDegrees?: number,`
- Fix `searchSites` createdAt and updatedAt filters to actually consider the provided date ranges

## NOTE
This PR has dependencies on these two, so please review them first if you haven't:

- #234
- #235

## Demo:
image

---

Thanks for the PR!

- [Backend](https://nr-site-registry-240-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-240-frontend.apps.silver.devops.gov.bc.ca)


frontend/src/app/features/site/dto/SiteFilterConfig.ts
anton-bcgov, line 1: Github didn't format this diff all that well, but the only changes here are:
- Put field definitions in a map-like structure for easier lookup by key
- Combine those definitions in an array with the same order as before.

backend/src/app/services/site/site.service.ts
```diff
@@ -254,14 +258,16 @@ export class SiteService {
...
+        start: whenCreated[0],
+        end: whenCreated[1],
       });
     }
 
     if (whenUpdated) {
-      query.andWhere('sites.whenUpdated = :whenUpdated', {
-        whenUpdated: whenUpdated,
+      query.andWhere('sites.whenUpdated BETWEEN :start AND :end', {
+        start: whenUpdated[0],
```
midhun-aot, line 277: Please add  a validation for array length before accessing array index

anton-bcgov: done: https://github.com/bcgov/nr-site-registry/pull/240/commits/2e69bc7eb0a0c127a5e28e51e16e06e3d7b9da68

nupurdixit13 (approved):

PR #258, opened 2025-01-14
feat: SRS-555 Map radius search

nupurdixit13:
# Description

These changes show site markers along with the results drawer for the sites returned on Radius Search.

---

Thanks for the PR!

- [Backend](https://nr-site-registry-258-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-258-frontend.apps.silver.devops.gov.bc.ca)


nikhila-aot (approved):

frontend/src/app/features/map/CircleLayer.tsx
```diff
@@ -1,22 +1,37 @@
...
   useMapCrosshairsCursor(map);
 
+  const initialSitesRef = useRef<Site[]>([]);
+
+  useEffect(() => {
+    if (sites.length > 0 && initialSitesRef.current.length === 0) {
+      initialSitesRef.current = sites;
+    }
+  }, [sites]);
+
```
anton-bcgov, line 34: This assumes that all sites are present in the result of the initial request. This is not always the case: for example, when you directly open the map route (or just refresh the page) with query params: `/map?search=vernon`

frontend/src/app/features/map/CircleLayer.tsx
```diff
@@ -25,7 +40,83 @@ export function CircleLayer({
...
+        site.latdeg <= maxLat &&
+        site.longdeg >= minLon &&
+        site.longdeg <= maxLon
+      );
+    });
+
+    const sitesWithinCircle = sitesWithinBoundingBox.filter((site) => {
+      const distance =
+        site.latdeg !== null && site.longdeg !== null
+          ? getDistance(
```
anton-bcgov, line 93: We talked about this in private, but posting here for posterity. I'm not 100% certain that running these calculations on the frontend is the best approach here. I'll gather my thought on exactly why I think so by tomorrow's standup, let's chat with the rest of the team before we proceed.

frontend/src/app/features/map/CircleLayer.tsx
nupurdixit13: I want the bounding box always to calculate based on all the actual sites present. I initially had the concept of keeping the value of previousRef but then on rethinking it appeared incorrect to me. Thats why kept it to be always the actual number of sites. Let me know if I answered your question or feel free to have an offline discussion over chat or call.

frontend/src/app/features/map/CircleLayer.tsx
nupurdixit13: Sure thing, lets discuss this tomorrow.

frontend/src/app/features/map/CircleLayer.tsx
nupurdixit13: Well this is something which should not occur as all the searches should ideally be mutually exclusive. Polygon and Radius Search are already mutually exclusive as I already took care of it. Text Search and other Searches should also be not performed in parallel. Let me know your thoughts.

anton-bcgov (dismissed):
Ok, I gathered my thoughts about this over the night and I'm certain that we should move the radius search (and polygon search) logic to the backend, as a part of the `mapSearch` query.

My main concern with searching on the frontend is that this solution hinges on all sites being loaded into local state. This works right now, but I fully expect that we may need to change how the initial data is loaded to account for either poor database performance or slow rendering on low-power devices. For that, we may need to limit how much data is loaded on the initial request. Meaning that if a site is not included in the initial response, it won't be included in the additional filtering applied by the radius search.

As for the thing that you brought up in DMs:
>bounding box will include a very small subset of the sites to calculate the distance for the sites to be included for RS. thats the main catch for keeping it performant

>SO we dont keep calculating the distance of all the sites each time

That's pretty much what `findSitesWithinCircle` is already doing, isn't it? With `initialSites.filter` and then with `sitesWithinBoundingBox.filter`

We can, however, leverage site's `geometry` data and the PostGIS extension that we already have enabled in the DB. This means we can use various spatial functions that are apparently very performant even for large data sets (as in - much larger than what we have in SRS). Here's a quick prototype I cooked up:

```typescript
const { minLat, minLong, maxLat, maxLong } = {
      minLat: 59.02642787065197,
      maxLat: 60.31245776711576,
      minLong: -127.15467478749466,
      maxLong: -124.60802052500537,
    };

    const res = await this.siteRepository
      .createQueryBuilder('site')
      .where(
        `ST_Within(
        ST_Transform(site.geometry, 3005),
        ST_Transform(ST_MakeEnvelope(:minLong, :minLat, :maxLong, :maxLat, 4326), 3005)
      )`,
        { minLat, minLong, maxLat, maxLong },
      )
      .getMany();
```

The lat/long values are hardcoded and represent an area near BC's north border, but the point still stands. Give it a go, try plugging in other max lat/lng values and let me know what you think. In addition to that, we only need to pass the center point coordinates and the radius as parameters and leave the calculation of the bounding circle to the backend.

acoard-aot:
100% agree with Anton's point.   Just to emphasize it, ideally the polygon/radius search should be happening in the database (as per Anton's example) and not in memory in the API.  When I've done this in the past, database was fastest, then API, then lastly frontend.

nupurdixit13:
> Ok, I gathered my thoughts about this over the night and I'm certain that we should move the radius search (and polygon search) logic to the backend, as a part of the `mapSearch` query.
> 
> My main concern with searching on the frontend is that this solution hinges on all sites being loaded into local state. This works right now, but I fully expect that we may need to change how the initial data is loaded to account for either poor database performance or slow rendering on low-power devices. For that, we may need to limit how much data is loaded on the initial request. Meaning that if a site is not included in the initial response, it won't be included in the additional filtering applied by the radius search.
> 
> As for the thing that you brought up in DMs:
> 
> > bounding box will include a very small subset of the sites to calculate the distance for the sites to be included for RS. thats the main catch for keeping it performant
> 
> > SO we dont keep calculating the distance of all the sites each time
> 
> That's pretty much what `findSitesWithinCircle` is already doing, isn't it? With `initialSites.filter` and then with `sitesWithinBoundingBox.filter`
> 
> We can, however, leverage site's `geometry` data and the PostGIS extension that we already have enabled in the DB. This means we can use various spatial functions that are apparently very performant even for large data sets (as in - much larger than what we have in SRS). Here's a quick prototype I cooked up:
> 
> ```ts
> const { minLat, minLong, maxLat, maxLong } = {
>       minLat: 59.02642787065197,
>       maxLat: 60.31245776711576,
>       minLong: -127.15467478749466,
>       maxLong: -124.60802052500537,
>     };
> 
>     const res = await this.siteRepository
>       .createQueryBuilder('site')
>       .where(
>         `ST_Within(
>         ST_Transform(site.geometry, 3005),
>         ST_Transform(ST_MakeEnvelope(:minLong, :minLat, :maxLong, :maxLat, 4326), 3005)
>       )`,
>         { minLat, minLong, maxLat, maxLong },
>       )
>       .getMany();
> ```
> 
> The lat/long values are hardcoded and represent an area near BC's north border, but the point still stands. Give it a go, try plugging in other max lat/lng values and let me know what you think. In addition to that, we only need to pass the center point coordinates and the radius as parameters and leave the calculation of the bounding circle to the backend.

I agree with your concerns but I also need to be sure of 2 things before going ahead with this: 
1) The geometry data for the site is correct. Are we sure on this part?
2) I found that we have geometry data present for 5280 sites out of total 5833 sites. Where will we get the rest of the data from?

We should ideally get the results from the below query and we don't need to go around the bounding box coordinates anymore IMO.
`SELECT *
FROM sites
WHERE ST_DWithin(
  geom,
  ST_SetSRID(ST_MakePoint(:longitude, :latitude), 4326),
  :radius
)`

anton-bcgov:
@nupurdixit13 

>The geometry data for the site is correct. Are we sure on this part?

When using the hard-coded values, the result of the query matched what I got with the FE-only / distance calculation approach. So I choose to believe that the geometry values are valid 🙂 

>I found that we have geometry data present for 5280 sites out of total 5833 sites. Where will we get the rest of the data from?

In the data that I have locally, all sites where the `geometry` values are null, the `latdeg` and `longdeg` values are also set to 0, so I think we can safely ignore those.

> We should ideally get the results from the below query and we don't need to go around the bounding box coordinates anymore IMO.
>SELECT * FROM sites WHERE ST_DWithin( geom, ST_SetSRID(ST_MakePoint(:longitude, :latitude), 4326), :radius )

I love it 🔥

acoard-aot:
Very minor improvement idea (if tricky feel free to skip):

The default radius could be a function of the current zoom level.  Basically if you're zoomed out far, the radius should be higher. If you're zoomed in close, the radius should be smaller.

(I mentioned this in meeting, just putting a record here)

acoard-aot:
Hey @nupurdixit13 - is this PR currently stable and good to demo? If so, would you be able to merge it in now and open a new PR to update the implementation to use database search?

Context: We've been asked to do a demo Monday.  

If you've any concerns let me know, but from what we saw in the demo today it looked good.

nupurdixit13:
> Hey @nupurdixit13 - is this PR currently stable and good to demo? If so, would you be able to merge it in now and open a new PR to update the implementation to use database search?
> 
> Context: We've been asked to do a demo Monday.
> 
> If you've any concerns let me know, but from what we saw in the demo today it looked good.

Sure, I will merge this one and open a new for the BE implementation. Thanks!

PR #267, opened 2025-01-20
fix: Fixing error while clicking on Notation tab

midhun-aot:
- Added null check to fix the error while clicking on Notation Tab

---

Thanks for the PR!

- [Backend](https://nr-site-registry-267-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-267-frontend.apps.silver.devops.gov.bc.ca)

### DIFF
--- a/frontend/codegen.ts
+++ b/frontend/codegen.ts
@@ -2,7 +2,7 @@ import { CodegenConfig } from '@graphql-codegen/cli';
 
 const config: CodegenConfig = {
   overwrite: true,
-  schema: 'http://localhost:4007/graphql',
+  schema: 'http://host.docker.internal:4007/graphql',
   documents: ['src/**/*.graphql'],
   generates: {
     './src/graphql/generated.ts': {
--- a/frontend/codegen.ts
+++ b/frontend/codegen.ts
@@ -2,7 +2,7 @@ import { CodegenConfig } from '@graphql-codegen/cli';
 
 const config: CodegenConfig = {
   overwrite: true,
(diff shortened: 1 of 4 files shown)


frontend/codegen.ts
```diff
@@ -2,7 +2,7 @@ import { CodegenConfig } from '@graphql-codegen/cli';
 
 const config: CodegenConfig = {
   overwrite: true,
-  schema: 'http://host.docker.internal:4007/graphql',
+  schema: 'http://localhost:4007/graphql',
```
bolny: **thought**: this is interesting. I think this is why codegen always fails for me because I don't use Docker Desktop, and don't have it's host routing. @anton-bcgov can you comment on whether this will have any unintended consequences? I think that this is *probably* fine.

bolny (commented):
Changes look good to me. I'd like comment on whether or not the change to the codegen url will cause any issues, but I don't see any other reason to not ship it as is.

frontend/codegen.ts
midhun-aot: I forgot to revert it, we can check for a way to take this URL from .env if possible

bolny:
Also you might need to rebase your work because it looks like the PR build step is failing due to a change Anton merged earlier.

frontend/codegen.ts
midhun-aot: I have reverted this change now.

bolny (approved):
Looks good.

PR #271, opened 2025-01-21
fix: sonar

mishraomp:
# Description

Sonar setup was done previously but commented out as it did not work, retrying to see if it can work now.

---

Thanks for the PR!

Deployments, as required, will be available below:
- [Backend](https://nr-site-registry-271-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-271-frontend.apps.silver.devops.gov.bc.ca)

Please create PRs in draft mode.  Mark as ready to enable:
- [Analysis Workflow](https://github.com/bcgov/nr-site-registry/actions/workflows/analysis.yml)

After merge, new images are deployed in:
- [Merge Workflow](https://github.com/bcgov/nr-site-registry/actions/workflows/merge.yml)

PR #277, opened 2025-01-24
feat: Radius search on Maps 

nupurdixit13:
# Description

This is the Radius Search implementation using the logic to fetch sites from backend. The changes also includes the results available on page reload. 
I have also introduced a new feature by disabling `Set Radius` for the first time as it doesn't make sense to set radius prior to selecting center. Will discuss this with @harry on Monday.

This PR does not include the below features:
- Mutual exclusivity between different searches

Thanks!
---

Thanks for the PR!

- [Backend](https://nr-site-registry-277-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-277-frontend.apps.silver.devops.gov.bc.ca)

---

Thanks for the PR!

- [Backend](https://nr-site-registry-277-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-277-frontend.apps.silver.devops.gov.bc.ca)

---

Thanks for the PR!

- [Backend](https://nr-site-registry-277-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-277-frontend.apps.silver.devops.gov.bc.ca)

---

Thanks for the PR!

- [Backend](https://nr-site-registry-277-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-277-frontend.apps.silver.devops.gov.bc.ca)

---

Thanks for the PR!

- [Backend](https://nr-site-registry-277-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-277-frontend.apps.silver.devops.gov.bc.ca)


frontend/src/app/features/map/CircleLayer.tsx
anton-bcgov, line 1: thanks for deleting this. I always would get confused over which component we're actually using: this one or `RadiusSearchLayer.tsx`

backend/src/app/services/site/site.service.ts
```diff
@@ -393,8 +396,20 @@ export class SiteService {
...
+      const [latitude, longitude] = center;
+
+      query.where(
+        `ST_DWithin(
+          ST_Transform(sites.geometry, ${BC_ALBERS}),
+          ST_Transform(ST_SetSRID(ST_MakePoint(${longitude}, ${latitude}),${WGS_84}), ${BC_ALBERS}),
+          ${radius}
+          )`,
+      );
+    }
```
anton-bcgov, line 440: Could you also add a test for this block? Verifying that `ST_MakePoint` is called with the correct arguments should be fine.

frontend/src/app/features/map/search/RadiusSearch.tsx
```diff
@@ -72,21 +70,20 @@ export function RadiusSearch({
...
+      <>
+        <Button size="medium" onClick={onCancelRadiusSearch}>
+          <XmarkIcon />
+          Cancel
+        </Button>
+        <DropdownButton
+          id="pointSearchSetRadiusButton"
+          menuClassName="point-search-menu"
+          dropdownContent={sliderBox}
+          disabled={!isCenterValid}
```
anton-bcgov, line 90: nice 👍

frontend/src/app/features/map/MapView.tsx
```diff
@@ -37,13 +36,20 @@ function MapView() {
   // Feature flag for turning OpenStreetMap tiles gray
   const osmGrayscale = false;
 
-  const { searchTerm, activeTool, polygonVertices } = useMapSearchContext();
+  const { searchTerm, activeTool, polygonVertices, center, radius } =
+    useMapSearchContext();
+
+  const variables: any = {
```
anton-bcgov: this should be
```typescript
const variables: MapSearchQueryVariables = { 
```

You can import this type from `generated.ts`

frontend/src/app/features/map/search/RadiusSearchButton.tsx
```diff
@@ -1,13 +1,15 @@
...
 import { MapPin } from '../../../components/common/icon';
+import { useMapSearchContext } from '../mapSearchContext/MapSearchContext';
 
 interface Props {
   isActive: boolean;
   onClick: () => void;
 }
 
 export function RadiusSearchButton({ isActive, onClick }: Readonly<Props>) {
+  const { handleRadiusToolClick } = useMapSearchContext();
```
anton-bcgov: Looks like this isn't used. Let's remove either `handleRadiusToolClick` or the `onClick` prop.

frontend/src/app/features/map/mapSearchContext/MapSearchContext.tsx
```diff
@@ -115,6 +139,41 @@ export const MapSearchQueryProvider = ({
 
   const clearQuery = () => setQuery({}, 'replace');
 
+  const onRadiusCrossHairClick = (newCenter: LatLngTuple) => {
```
anton-bcgov: I suggest renaming it to `setCircleCenter`.

`onRadiusCrossHairClick` - describes when it's supposed to happen
`setCircleCenter` - describes what it actually does

Since this function is only provided by the context and the consumer decides _when_ it should fire, I think the second option is more appropriate for the name.

(I also fully admit that I don't always follow this practice :P)

frontend/src/app/features/map/mapSearchContext/MapSearchContext.tsx
```diff
@@ -115,6 +139,41 @@ export const MapSearchQueryProvider = ({
...
+        'replace',
+      );
+    }
+  };
+
+  const onCancelRadiusSearch = () => {
+    clearQuery();
+    setActiveToolState(null);
+    setRadius(MIN_CIRCLE_RADIUS);
+  };
```
anton-bcgov: Same comment about the names of these two functions. Let's describe _what_ they do, not _when_ they do it.

frontend/src/app/features/map/mapSearchContext/MapSearchContext.tsx
```diff
@@ -115,6 +139,41 @@ export const MapSearchQueryProvider = ({
...
 
+  const onRadiusCrossHairClick = (newCenter: LatLngTuple) => {
+    setCenter(newCenter);
+    setQuery(
+      { circle: { center: newCenter, radius: radius || MIN_CIRCLE_RADIUS } },
+      'replace',
+    );
+  };
+
+  const onRadiusChange = (_ev: any, value: number | number[]) => {
```
anton-bcgov: If we don't care about the `_env` parameter, let's not pass it here at all?

I also think that the `value` parameter should just be `number`. Passing arguments of proper type should be a responsibility of the component that calls this function.

frontend/src/app/features/map/search/RadiusSearch.tsx
```diff
@@ -59,8 +56,9 @@ export function RadiusSearch({
...
         step={MIN_CIRCLE_RADIUS}
         defaultValue={MIN_CIRCLE_RADIUS}
-        value={radius}
-        onChange={onRadiusChange}
+        value={localRadius}
+        onChangeCommitted={handleChangeCommitted} //Make fetch request on change commit
+        onChange={handleChange} //This is kept to keep the slider move smoothly without making fetch requests
       />
       <Typography className="point-search-slider-text">
         {formatDistance(radius, 1)}
```
anton-bcgov: Let's change this to `{formatDistance(localRadius, 1)}`. It's a bit more responsive from the point of UX.

frontend/src/app/features/map/layers/RadiusSearchLayer.tsx
```diff
@@ -1,29 +1,50 @@
...
   onCrossHairClick,
-  sites,
-  setSites,
 }: Readonly<RadiusSearchLayerProps>) {
-  const { activeTool } = useMapSearchContext();
+  const { center, radius, onRadiusCrossHairClick, activeTool } =
+    useMapSearchContext();
+  const map = useMap();
+
+  useMapCrosshairsCursor(map);
```
anton-bcgov: This causes the `CrosshairsCursor` to be visible when the map first loads, even without the circle query param.

Try changing it to this:
```suggestion
  useMapCrosshairsCursor(map, activeTool === ActiveToolEnum.radiusSearch);
```

frontend/src/app/features/map/layers/RadiusSearchLayer.tsx
```diff
@@ -1,29 +1,50 @@
...
+
+  useMapEvents({
+    click: (ev: LeafletMouseEvent) => {
+      const newCenter: LatLngTuple = [ev.latlng.lat, ev.latlng.lng];
+      onRadiusCrossHairClick(newCenter);
+      onCrossHairClick();
+    },
+  });
+  const drawCircle = center && radius > 500;
```
anton-bcgov: Should the 500 here be `MIN_CIRCLE_RADIUS`?

anton-bcgov (commented):
Thanks for working on this Nupur, this looks great so far!

I found one weird edge-casey bug though, not sure what causes it. Steps to reproduce:
1. Open RS tool, set center and radius, and refresh the page (or just navigate to the map route with set query params like `/map?circle=%7B"center"%3A%5B54.939765758658936%2C-128.40820312500003%5D%2C"radius"%3A197000%7D`
2. Click anywhere on the map - this causes the radius to reset (or set to some small value, I'm not sure) and eventually the page loads all site markers.


Demo: 


https://github.com/user-attachments/assets/988f3451-7e23-4426-9497-574a93f339bf

nupurdixit13:
> Thanks for working on this Nupur, this looks great so far!
> 
> I found one weird edge-casey bug though, not sure what causes it. Steps to reproduce:
> 
> 1. Open RS tool, set center and radius, and refresh the page (or just navigate to the map route with set query params like `/map?circle=%7B"center"%3A%5B54.939765758658936%2C-128.40820312500003%5D%2C"radius"%3A197000%7D`
> 2. Click anywhere on the map - this causes the radius to reset (or set to some small value, I'm not sure) and eventually the page loads all site markers.
> 
> Demo:
> 
>  Screen.Recording.2025-01-24.at.7.08.50.PM.mov

I see the reason behind it. Thanks for bringing this up.

backend/src/app/services/site/site.service.ts
nupurdixit13: Oh yes, Unit tests are still pending. Will do them in a separate PR.

frontend/src/app/features/map/search/RadiusSearchButton.tsx
nupurdixit13: I have removed the unused `handleRadiusToolClick`

frontend/src/app/features/map/mapSearchContext/MapSearchContext.tsx
nupurdixit13: If the center would have been set by multiple ways then giving it a generic name would have been fine but here the trigger is just one. Will thus rename it to `setCenterOnCrossHairClick`. This makes it clear what the function does and when it happens. Let me know if that sounds fine :)  
Ideally the name should convey both the intent (what it does) and the trigger or context (when or why it happens), if applicable.

frontend/src/app/features/map/mapSearchContext/MapSearchContext.tsx
nupurdixit13: Will rename `onCancelRadiusSearch` to `clearRadiusSearch` and `onRadiusChange` to `handleRadiusChange`

frontend/src/app/features/map/mapSearchContext/MapSearchContext.tsx
nupurdixit13: Its mandatory to pass it as the definitions is ` onChangeCommitted?: (event: React.SyntheticEvent | Event, value: number | number[]) => void;`

nupurdixit13:
> > Thanks for working on this Nupur, this looks great so far!
> > I found one weird edge-casey bug though, not sure what causes it. Steps to reproduce:
> > 
> > 1. Open RS tool, set center and radius, and refresh the page (or just navigate to the map route with set query params like `/map?circle=%7B"center"%3A%5B54.939765758658936%2C-128.40820312500003%5D%2C"radius"%3A197000%7D`
> > 2. Click anywhere on the map - this causes the radius to reset (or set to some small value, I'm not sure) and eventually the page loads all site markers.
> > 
> > Demo:
> > Screen.Recording.2025-01-24.at.7.08.50.PM.mov
> 
> I see the reason behind it. Thanks for bringing this up. and yes, thanks for the detailed review and comments. Super helpful!

bolny:
I looks like two checks are failing due to lack of unit test coverage. We need to make a decision about whether or not we will enforce this and change the requirement with another PR if we decide to not enforce test coverage. The mandatory coverage was implemented in https://github.com/bcgov/nr-site-registry/pull/271,

backend/src/app/services/site/site.service.ts
anton-bcgov: Now that we have required code coverage checks, we'll need to do it in one go. We will likely make FE coverage optional for now, but BE coverage will remain required.

Or feel free to add a separate PR with this one as the base.

frontend/src/app/features/map/mapSearchContext/MapSearchContext.tsx
anton-bcgov: Yes, the callback passed to `onChangeCommitted` prop need to have the event, but we don't need it in the context.

In `RadiusSearch.tsx` we already have
```typescript
const handleChangeCommitted = (_ev: any, newValue: number | number[]) => {
    handleRadiusChange(_ev, newValue);
};
```

You can move the value extraction logic here (this is pretty much a copy of the code you already had in `handleChange` in the same file) :

```typescript
const handleChangeCommitted = (_ev: any, newValue: number | number[]) => {
    if (typeof newValue === 'number') {
      handleRadiusChange(newValue);
    } else if (Array.isArray(newValue) && newValue.length > 0) {
      handleRadiusChange(newValue[0]);
    }
  };
```

backend/src/app/services/site/site.service.ts
nupurdixit13: Sure, will add UT's in this PR today

frontend/src/app/features/map/mapSearchContext/MapSearchContext.tsx
nupurdixit13: I see what you mean, yeah that should be fine too.

backend/src/app/services/site/site.service.ts
```diff
@@ -396,6 +397,35 @@ export class SiteService {
...
+      if (
+        latitude === 0 ||
+        latitude === undefined ||
+        longitude === 0 ||
+        longitude === undefined
+      ) {
+        throw new HttpException(
+          'Latitude and longitude cannot be 0 or undefined',
+          HttpStatus.BAD_REQUEST,
+        );
```
anton-bcgov: I know that this is highly unlikely that someone would do this in SRS, but `0,0` are valid coordinates. I don't see a reason to throw an error here.

frontend/src/app/features/map/layers/RadiusSearchLayer.tsx
```diff
@@ -27,7 +28,12 @@ export function RadiusSearchLayer({
       onCrossHairClick();
     },
   });
-  const drawCircle = center && radius > MIN_CIRCLE_RADIUS;
+
+  //displayRadius is only set to circle with MIN_CIRCLE_RADIUS so that it is visible on the map with a clear dot/circle
+  const displayRadius =
```
anton-bcgov: Did you mean to use this variable as a prop to `<Circle>`? It is unused currently

anton-bcgov: Actually, let's leave this logic out of the scope of this PR. I think the calculation of the initial radius based on zoom needs to be a bit more sophisticated.

anton-bcgov (commented):
Left a few more comments, but it's getting really close.

Also, not sure if this is a new bug:


https://github.com/user-attachments/assets/271759f9-5342-4f76-8cdd-41b2ace5adbb

nupurdixit13:
> Left a few more comments, but it's getting really close.
> 
> Also, not sure if this is a new bug:
> 
>  Screen.Recording.2025-01-28.at.5.21.36.PM.mov

Interesting, need to deep dive into it.

frontend/src/app/features/map/layers/RadiusSearchLayer.tsx
nupurdixit13: Yeah, the initial dot appearing without this logic should be good enough IMO.

backend/src/app/services/site/site.service.ts
nupurdixit13: Hmm. I see...I will add a check for null which I anyways missed since just checked for undefined

frontend/src/app/features/map/layers/RadiusSearchLayer.tsx
nupurdixit13: > Yeah, the initial dot appearing without this logic should be good enough IMO. Anyways it was only catering the Min radius.

nupurdixit13: > Did you mean to use this variable as a prop to `<Circle>`? It is unused currently

Yeah that was the idea , I guess I was testing something else and forgot to change it back to `radius={displayRadius}`. But I guess based on your comment below will remove it altogether

anton-bcgov (dismissed):
Thanks for all the iterations on this! This is looking 🤌

midhun-aot (approved):

nikhila-aot (approved):

PR #327, opened 2025-06-27
fix: Replace entity with DTO

midhun-aot:
- Replacing Site Entity With DTO

---

Thanks for the PR!

- [Backend](https://nr-site-registry-327-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-327-frontend.apps.silver.devops.gov.bc.ca)


backend/src/app/dto/response/genericResponse.ts
```diff
@@ -57,6 +58,22 @@ export class FetchSiteDetail extends BaseHttpResponse {
   }
 }
 
+@ObjectType()
+export class FetchSiteDetailDTO extends BaseHttpResponse {
```
jaspalsingh-aot: Would suggest to have a class name with response instead of dto for example SiteDetailResponse

jaspalsingh-aot (changes requested):

backend/src/app/dto/site.dto.ts
```diff
@@ -0,0 +1,109 @@
+import { Field, ObjectType } from '@nestjs/graphql';
+
+@ObjectType()
+export class SiteDto {
```
jaspalsingh-aot: I noticed that the same set of properties is defined in both the input and object DTOs. While this works, it could lead to duplication and inconsistencies over time. I’d recommend extracting the shared properties into a common DTO for better reusability. You can refer to the CATS where we addressed a similar case. Also, having a clear folder structure would help improve readability.

jaspalsingh-aot (changes requested):

backend/src/app/dto/site.dto.ts
midhun-aot: update the code to remove duplicate and promote reuse

jaspalsingh-aot (approved):
LGTM

PR #440, opened 2026-01-13
chore: Update bcgov/action-test-and-analyse to latest version to support nod…

nupurdixit13:
# Description

We're using this action https://github.com/bcgov/action-test-and-analyse with a pinned 1.2.1 version in analysis.yml. The current version is 1.7.0 which runs a currently supported version of ubuntu and node. Thus updating the same.

---

Thanks for the PR!

- [Backend](https://nr-site-registry-440-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-440-frontend.apps.silver.devops.gov.bc.ca)

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -43,7 +43,7 @@ jobs:
           - dir: frontend
             token: SONAR_TOKEN_FRONTEND
     steps:
-      - uses: bcgov/action-test-and-analyse@v1.7.0
+      - uses: bcgov/action-test-and-analyse@v1.6.0
         env:
           LOG_LEVEL: info
           LOG_FILE_LOCATION: /tmp/sites-%DATE%.log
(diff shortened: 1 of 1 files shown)


.github/workflows/analysis.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           - dir: frontend
             token: SONAR_TOKEN_FRONTEND
     steps:
-      - uses: bcgov/action-test-and-analyse@v1.2.1
+      - uses: bcgov/action-test-and-analyse@v1.7.0
```
bolny: It looks like the latest version is 1.6.0. I probably just hit the wrong key when I typed it out.

```suggestion
      - uses: bcgov/action-test-and-analyse@v1.6.0
```

bolny: They actually release 2.0 today, so we might want to look into upgrading again, but there are likely breaking changes.

bolny (approved):
All systems green.

PR #469, opened 2026-03-17
fix: SRS-1460 default value to BC for province

urmikataria:
default value to BC for province

---

Thanks for the PR!

- [Backend](https://nr-site-registry-469-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-469-frontend.apps.silver.devops.gov.bc.ca)


nupurdixit13:
I tried the PR env link and found two issues
1. Looks like the field still has the asterisk which should be removed as it is no more mandatory field.
2. On saving the details, we get error as attached 
Screenshot 2026-03-17 at 10 16 18 AM
 I think it needs to be handled as well.

nupurdixit13 (commented):
Left few issues in the "Conversation" section

nupurdixit13 (approved):

nupurdixit13 (commented):
LGTM

PR #472, opened 2026-03-18
fix: Show the selected site on Map

nupurdixit13:
#Issue
When you click “View on Map” for a site:

it navigates to Map Search and opens the details drawer for that siteId
but the site doesn’t appear on the map (no marker/zoom), so the drawer is showing a site that isn’t represented visually.

#Root cause
The map pins come from the mapSearch query (returns a list of sites to render as markers).
The selected site in the drawer comes from URL state (?site=<id>) and triggers findSiteBySiteId to load drawer details.
If the selected site is not included in the current mapSearch results (common when arriving from a details page, or when map search params don’t include that site), SiteMarkers never receives it → no marker to highlight or zoom to.

#Fix 
Ensure the selected site is present in the marker list:
- In MapView.tsx, when selectedSiteId exists, fetch it via findSiteBySiteId and merge that site into the sites list used for markers (e.g., sitesToShow).
- Auto-pan/zoom to the selected site once it loads:
- Add an effect in MapView.tsx that calls flyTo/fitBounds when the selected site’s lat/long becomes available.
- Make the selected marker reliably “yellow” even with clustering:
- 
Before Fix:
Before

After fix:
After

In SiteMarkers.tsx, compare IDs using String(site.id) === String(selectedSiteId) (API may return numeric IDs).
Render the selected marker outside MarkerClusterGroup so it can’t be hidden inside a cluster icon.

---

Thanks for the PR!

- [Backend](https://nr-site-registry-472-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-472-frontend.apps.silver.devops.gov.bc.ca)

---

Thanks for the PR!

- [Backend](https://nr-site-registry-472-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-472-frontend.apps.silver.devops.gov.bc.ca)

### DIFF
--- a/frontend/src/app/features/map/MapView.test.tsx
+++ b/frontend/src/app/features/map/MapView.test.tsx
@@ -1,0 +1,237 @@
+import React from 'react';
+import { render, screen, waitFor } from '@testing-library/react';
+import type { LatLngTuple, Map } from 'leaflet';
+
+const mockLeafletMap: Partial<Map> = {
+  flyToBounds: jest.fn(),
+  flyTo: jest.fn(),
+} as Partial<Map>;
+
+const siteData = [
+  {
+    id: 's1',
+    addrLine_1: '123',
+    latdeg: 49,
+    longdeg: -123,
+  },
+];
+
+jest.mock('react-leaflet', () => {
+  const React = require('react') as typeof import('react');
+  return {
+    __esModule: true,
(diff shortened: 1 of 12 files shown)


bolny (approved):
I can't find anything that I would change, and it works well for me!

PR #487, opened 2026-03-31
fix: Update migration sequence default value.

bolny:
Update migration to default to a sequence value of 1 if there are no ids. 0 is an invalid value.

1. `GREATEST(..., 1)` ensures the value passed to `setval` is never `0`, which PostgreSQL rejects.
2. `is_called` flag; `(SELECT COUNT(*) > 0 FROM sites.subdivisions)` evaluates to false when empty, telling PostgreSQL the sequence hasn't been used yet, so the next `nextval` returns `1` rather than `2`. Not a big deal, but I like it.

---

Thanks for the PR!

- [Backend](https://nr-site-registry-487-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-487-frontend.apps.silver.devops.gov.bc.ca)

### DIFF
--- a/charts/crunchy/values.yaml
+++ b/charts/crunchy/values.yaml
@@ -1,4 +1,4 @@
-global:  # Trigger a db deploy.
+global:
   config:
     dbName: site #test
 crunchy: # enable it for TEST and PROD, for PR based pipelines simply use single postgres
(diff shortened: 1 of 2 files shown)


charts/crunchy/values.yaml
```diff
@@ -1,4 +1,4 @@
-global:
+global:  # Trigger a db deploy.
```
bolny:
```suggestion
global:
```

urmikataria (approved):

PR #492, opened 2026-04-07
fix: SRS-1432

urmikataria:
---

Thanks for the PR!

- [Backend](https://nr-site-registry-492-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-492-frontend.apps.silver.devops.gov.bc.ca)

### DIFF
--- a/frontend/src/app/features/details/documents/Documents.tsx
+++ b/frontend/src/app/features/details/documents/Documents.tsx
@@ -742,8 +742,8 @@ const Documents: React.FC<IComponentProps> = ({ showPending = false }) => {
       <Alert variant={hasDocuments ? 'info' : 'warning'} data-testid="no-site">
         {hasDocuments
           ? userType === UserType.Internal ?
-            'No documents found for this site. To inquire about documents, submit a Site Information Request form, or contact Advisor.SiteInformation@gov.bc.ca' :
-            'No documents found for this site. Please add documents to it.'
+            'No documents found for this site.' :
+            'No documents found for this site. To inquire about documents, submit a Site Information Request form, or contact Advisor.SiteInformation@gov.bc.ca'
             : 'Please create a site before adding documents. Once the site is created, you can add documents to it.'
         }
       </Alert>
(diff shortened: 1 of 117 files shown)


frontend/src/app/features/details/documents/Documents.tsx
```diff
@@ -761,10 +761,11 @@ const Documents: React.FC<IComponentProps> = ({ showPending = false }) => {
...
       <Alert variant={hasDocuments ? 'info' : 'warning'} data-testid="no-site">
         {hasDocuments
-          ? userType === UserType.Internal
-            ? 'No documents found for this site. To inquire about documents, submit a Site Information Request form, or contact Advisor.SiteInformation@gov.bc.ca'
-            : 'No documents found for this site. Please add documents to it.'
-          : 'Please create a site before adding documents. Once the site is created, you can add documents to it.'}
+          ? userType === UserType.Internal ?
+            'No documents found for this site.' :
+            'No documents found for this site. To inquire about documents, submit a Site Information Request form, or contact Advisor.SiteInformation@gov.bc.ca'
+            : 'Please create a site before adding documents. Once the site is created, you can add documents to it.'
```
bolny: Non-blocking nitpick: If we're going to use nested ternary operators this way, I'd prefer to keep them consistently at the front of the line or at the end of the line, but not mixed.


```suggestion
          ? userType === UserType.Internal
            ? 'No documents found for this site.'
            : 'No documents found for this site. To inquire about documents, submit a Site Information Request form, or contact Advisor.SiteInformation@gov.bc.ca'
          : 'Please create a site before adding documents. Once the site is created, you can add documents to it.'
```

bolny (dismissed):
The logic looks like it is correct to me.

At some point I'd like to have a conversation with the team about writing conditional copy in templates like this. We are forced to use ternary operators and I strongly suspect that the original logic error was caused because they are difficult to read. I think we should be pulling this logic out into a function call with easy to read if-else syntax.

bolny (approved):

PR #495, opened 2026-04-09
fix: Hide private sites from external and unaunthenticated users

nupurdixit13:
# Description (what this PR fixes)
This PR fixes the map discovery bug where private sites were visible to external (non‑IDIR) users on the map.

# Issue
The main registry search (searchSites) already restricts non‑IDIR / anonymous users to sr_action = public sites.
The map endpoints did not consistently apply the same rule:
mapSearch (pins: initial load, text search, polygon, radius) could return private sites.
findSitesAndPlaces (map autocomplete) could suggest private sites.

# Fix
mapSearch now applies the same disclosure filter as searchSites for non‑IDIR/anonymous users:
non‑IDIR / anonymous → only sr_action = public sites returned as pins/search results
IDIR → unchanged
findSitesAndPlaces now applies the same disclosure filter to the sites portion of autocomplete results for non‑IDIR/anonymous users (places unaffected).
This is implemented via a shared helper (applyPublicVisibilityFilterForNonIdirUsers) so the rule stays consistent across these “discovery/list” endpoints.

# Test coverage
Updated/added unit tests to verify:
mapSearch applies sr_action = public for non‑IDIR users and does not for IDIR.
findSitesAndPlaces applies the same sr_action = public restriction for non‑IDIR (sites list)

# Important note / known gap (not addressed in this PR)
Direct navigation by site ID is still possible because findSiteBySiteId currently fetches by { id } without enforcing public-only visibility for external/anonymous users:

If a site row is still in the DB with sr_action = private (or any state treated as non‑public), an external or anonymous user can still open /site/details/:id or /map?site=:id and potentially get full details, because findSiteBySiteId loads by id only (no public-only filter on that path for the normal non‑pending case).

Separate ticket: apply the same disclosure rule to findSiteBySiteId (details-by-id) and optionally implement a silent redirect to search when details are unavailable.

---

Thanks for the PR!

- [Backend](https://nr-site-registry-495-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-495-frontend.apps.silver.devops.gov.bc.ca)

### DIFF
--- a/backend/src/app/services/site/site.service.ts
+++ b/backend/src/app/services/site/site.service.ts
@@ -403,10 +403,12 @@ export class SiteService {
     searchTerm,
     polygon,
     circle,
+    userInfo,
   }: {
     searchTerm?: string;
     polygon?: LatLngTuple[];
     circle?: RadiusSearchParams;
+    userInfo?: any;
   }) {
     this.sitesLogger.log('SiteService.mapSearch() start');
 
@@ -507,12 +509,14 @@ export class SiteService {
       );
     }
 
+    this.applyPublicVisibilityFilterForNonIdirUsers(query, userInfo);
+
     const [result] = await query.getManyAndCount();
     this.sitesLogger.log('SiteService.mapSearch() end');
     return result;
   }
 
-  async findSitesAndPlaces(searchTerm = '', limit = 3) {
+  async findSitesAndPlaces(searchTerm = '', limit = 3, userInfo?: any) {
     this.sitesLogger.log('SiteService.findSitesAndPlaces() start');
 
     const searchTermClean = searchTerm.toLowerCase().trim();
@@ -543,6 +547,8 @@ export class SiteService {
         'ASC',
       )
       .addOrderBy('sites.id', 'ASC');
+
+    this.applyPublicVisibilityFilterForNonIdirUsers(sitesQuery, userInfo);
     placesQuery
       .where('LOWER(places.name) LIKE LOWER(:searchTerm)', {
         searchTerm: `%${searchTermClean}%`,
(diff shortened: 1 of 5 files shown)


bolny (dismissed):
Looking good!

backend/src/app/services/site/site.service.ts
```diff
@@ -558,6 +564,19 @@ export class SiteService {
     return { sites, places };
   }
 
+  private applyPublicVisibilityFilterForNonIdirUsers(
+    query: any,
+    userInfo?: any,
+  ) {
+    // `sr_action` is a disclosure/visibility flag. Non-IDIR users must never receive non-public sites
+    // (e.g., private sites) from "list-style" endpoints like map pins or autocomplete.
+    if (!userInfo || userInfo?.identity_provider !== UserTypeEum.IDIR) {
```
urmikataria, line 573: if (!userInfo || userInfo.identity_provider !== UserTypeEnum.IDIR) {

we can consider userInfo is present if first case doesnt pass of !userInfo

urmikataria (dismissed):
just one quick suggestion else all looks good

backend/src/app/services/site/site.service.ts
nupurdixit13: > if (!userInfo || userInfo.identity_provider !== UserTypeEnum.IDIR) {
> 
> we can consider userInfo is present if first case doesnt pass of !userInfo

I agree, optional thing could be skipped and I can safely say "if (!userInfo || userInfo.identity_provider !== UserTypeEum.IDIR) {". 
I hope thats what you meant.

bolny:
I think the sonarqube coverage check is failing because the specs don't cover the case where userinfo is null.

backend/src/app/services/site/site.service.ts
urmikataria: yes exactly

nupurdixit13:
> I think the sonarqube coverage check is failing because the specs don't cover the case where userinfo is null.

In the workflow, the Sonar action runs with dir: backend / dir: frontend.
Backend Jest writes LCOV to backend/src/coverage/lcov.info (because rootDir: "src" and coverageDirectory: "./coverage").
The workflow previously pointed Sonar only to ./coverage/lcov.info, which doesn’t exist for backend.

My latest commit handles it and it should be good for all future PR's:
-Dsonar.javascript.lcov.reportPaths=./src/coverage/lcov.info,./coverage/lcov.info

urmikataria (approved):
LGTM

PR #501, opened 2026-04-14
fix: SRS-1517 Add character limit to fields to be consistent with db

urmikataria:
---

Thanks for the PR!

- [Backend](https://nr-site-registry-501-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-501-frontend.apps.silver.devops.gov.bc.ca)

### DIFF
--- a/frontend/src/app/components/input-controls/InputControls.tsx
+++ b/frontend/src/app/components/input-controls/InputControls.tsx
@@ -154,17 +154,21 @@ export const TextInput: React.FC<InputProps> = ({
   const [error, setError] = useState<string | null>(null);
 
   const validateInput = (inputValue: string) => {
-    if (validation) {
-      if (validation.required && !inputValue.trim()) {
-        setError(validation.customMessage || ' ');
-        return false;
-      }
-      if (validation?.pattern && !validation.pattern?.test(inputValue)) {
-        setError(validation.customMessage || '');
-        return false;
-      }
+    if (validation?.required && !inputValue.trim()) {
+      setError(validation.customMessage || ' ');
+      return false;
     }
-
+    if (validation?.pattern && !validation.pattern?.test(inputValue)) {
+      setError(validation.customMessage || '');
+      return false;
+    }
+    if (
+      validation?.maxLength &&
+      inputValue.length > validation.maxLength
+    ) {
+      setError(`Maximum ${validation.maxLength} characters allowed`);
+      return false;
+    }
     setError(null);
     return true;
   };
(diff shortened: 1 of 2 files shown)


frontend/src/app/components/input-controls/InputControls.tsx
```diff
@@ -154,32 +154,34 @@ export const TextInput: React.FC<InputProps> = ({
...
+    if (validation?.required && !inputValue.trim()) {
+      setError(validation.customMessage || ' ');
+      return false;
+    }
+    if (validation?.pattern && !validation.pattern?.test(inputValue)) {
+      setError(validation.customMessage || '');
+      return false;
+    }
+    if (
+      validation?.maxLength &&
```
jaspalsingh-aot, line 166: It looks like maxLength validation has been added within validateInput. Since maxLength is already supported at the input attribute level in the previous implementation, this may introduce some duplication in validation logic and could potentially lead to inconsistent behavior.

As a suggestion, we could consider continuing to rely on the existing maxLength and minLength input attributes to handle length restrictions. This would help keep the validation logic simpler and avoid redundancy while still providing a good user experience.

frontend/src/app/components/input-controls/InputControls.tsx
```diff
@@ -154,32 +154,34 @@ export const TextInput: React.FC<InputProps> = ({
...
-
+    
     if (allowNumbersOnly || (inputValue && !validateInput(inputValue))) {
       return;
     }
 
-    if (validation?.required) {
-      validateInput(inputValue);
-    }
+    validateInput(inputValue);
```
jaspalsingh-aot, line 184: Given that we already have maxLength handling implemented at the input attribute level, I would suggest keeping the existing logic as is. This helps avoid duplication and keeps the validation flow simpler and more consistent.

frontend/src/app/components/input-controls/InputControls.tsx
```diff
@@ -214,9 +216,6 @@ export const TextInput: React.FC<InputProps> = ({
           aria-label={label} // Accessibility
           required={error ? true : false}
           disabled={isDisabled ?? false}
-          {...(validation?.maxLength
```
jaspalsingh-aot, line 217: To keep things simple and consistent, I would suggest retaining the existing logic as it is. We already have support for restricting input length at the UI level using maxLength and minLength.

We may just need to ensure that maxLength is properly passed in the validation object so that users are prevented from entering more characters than allowed. This approach helps enforce limits effectively while avoiding additional or duplicate validation logic.

jaspalsingh-aot (changes requested):

jaspalsingh-aot:
To keep things simple and consistent, I would suggest retaining the existing logic as it is. We already have support for restricting input length at the UI level using maxLength and minLength.

We may just need to ensure that maxLength is properly passed in the validation object so that users are prevented from entering more characters than allowed. This approach helps enforce limits effectively while avoiding additional or duplicate validation logic.

jaspalsingh-aot (approved):
LGTM

PR #517, opened 2026-04-28
chore: Temporarily Disable PDB

bolny:
# Description

This disables our Pod Disruption Budget in Prod and Test. 

There is currently an issue where we can only have one backend pod running due to a race condition with the database. If we only have one pod with a minimum available of one, then it causes issues for operations because we can't scale down the service using a rolling restart for operations like infrastructure migrations.

A proper fix for this is being tracking in SRS-1559

Fixes # SRS-1469

---

Thanks for the PR!

- [Backend](https://nr-site-registry-517-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-517-frontend.apps.silver.devops.gov.bc.ca)

### DIFF
--- a/charts/app/values-prod.yaml
+++ b/charts/app/values-prod.yaml
@@ -9,8 +9,9 @@ global:
 
 backend:
   pdb:
-    enabled: true
-    # Trying to force replicaCount to 1 during deployment so we only run the seed data scripts once.
+    # Temporarily disabled until we can come up with a strategy for database seeding with multiple pods. See SRS-1559.
+    enabled: false
+
 
   # Temporarily going for 1 pod max, so that we only run the SEED_DATA import scripts once, instead of twice concurrently.
   replicaCount: 1
--- a/charts/app/values.yaml
+++ b/charts/app/values.yaml
@@ -68,7 +68,7 @@ backend:
       route.openshift.io/termination: "edge"
   pdb:
     # Temporarily disabled until we can come up with a strategy for database seeding with multiple pods. See SRS-1559.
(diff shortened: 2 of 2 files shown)


charts/app/values.yaml
```diff
@@ -67,7 +67,8 @@ backend:
     annotations:
       route.openshift.io/termination: "edge"
   pdb:
-    enabled: true # enable it in PRODUCTION for having pod disruption budget.
+    # Temporarily disabled until we can come up with a strategy for database seeding with multiple pods. See SRS-1559.
+    enabled: false # enable it in PRODUCTION for having pod disruption budget.
```
nupurdixit13: Comment should say "disable"

bolny (commented):
> Comment should say "disable" 

That comment is an instruction telling the user of QuickStart OpenShift that they should enable the PDB in production. I'm leaving it there because we are only temporarily disabling it until we can resolve this issue.

charts/app/values.yaml
nupurdixit13: This was also affect TEST, do you intend it as well?

charts/app/values-prod.yaml
```diff
@@ -9,8 +9,9 @@ global:
 
 backend:
   pdb:
-    enabled: true
```
nupurdixit13, line 12: Another option would be to add  : `minAvailable: 0` as you have already tried out before leaving enable to true. Just an option

nupurdixit13 (approved):

PR #520, opened 2026-05-01
fix: Split public vs logged-in findSiteBySiteId and remove pending from public

nupurdixit13:
The public “site by id” API accepted a pending  flag, so any client—including external users—could send pending: true and ask the server for pending / not-yet-approved site data. That is an authorization issue: the GraphQL contract made pending reads look like a normal, client-controlled switch instead of a staff-only capability. We removed pending from the public field and route pending review through findSiteBySiteIdLoggedInUser, with the service still enforcing who may treat a site as pending (e.g. IDIR), so externals can’t drive pending access by tweaking query variables.

The public GraphQL field findSiteBySiteId no longer accepts pending and always loads non-pending/public behaviour; findSiteBySiteIdLoggedInUser is used when the app needs pending and a real user context (logged-in and SR flows).

---

Thanks for the PR!

- [Backend](https://nr-site-registry-520-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-520-frontend.apps.silver.devops.gov.bc.ca)

### DIFF
--- a/backend/src/app/resolvers/site/sitePublic.resolver.spec.ts
+++ b/backend/src/app/resolvers/site/sitePublic.resolver.spec.ts
@@ -91,36 +91,33 @@ describe('SiteResolver', () => {
   describe('findSiteBySiteId', () => {
     it('should call siteService.findSiteBySiteId with the provided siteId', () => {
       const siteId = '123';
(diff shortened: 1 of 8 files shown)


midhun-aot (approved):
LGTM

PR #531, opened 2026-05-11
chore: force deploy

bolny:
Don't review this. I'm just using it to do a helm deploy so I can look at its configuration as it is applied to the openshift namespace.

---

Thanks for the PR!

- [Backend](https://nr-site-registry-531-backend.apps.silver.devops.gov.bc.ca)
- [Frontend](https://nr-site-registry-531-frontend.apps.silver.devops.gov.bc.ca)


nupurdixit13:
I dont see any change other than a comment, is that what you intended?

charts/crunchy/values.yaml
```diff
@@ -1,4 +1,5 @@
 global:
+  # Force db deploy
```
nupurdixit13, line 2: Is this the only change?

